PR #50, opened 2023-03-22
Add Google Storage Lock

veziak:


pkg/gcp/gcp_lock.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+type GoogleStorageLock struct {
+	Client *storage.Client
+	Bucket *storage.BucketHandle
+	ctx    context.Context
+}
+
+type Lock interface {
+	Lock(transactionId int, resource string) (bool, error)
```
Spartakovic: we should use existing interface

veziak: done

cmd/digger/main.go
```diff
@@ -1,36 +1,48 @@
...
 	}
-	sess := session.Must(session.NewSession())
-	dynamoDb := dynamodb.New(sess)
-	dynamoDbLock := aws.DynamoDbLock{DynamoDb: dynamoDb}
+	println("Digger config read successfully")
+
+	lock, err := utils.GetLock()
+	if err != nil {
+		fmt.Printf("Failed to create lock. %s\n", err)
+		os.Exit(2)
```
Spartakovic, line 24: why do we want different exit codes?

veziak: it was convenient when no error message has been displayed, not sure myself if we need to leave it this way or return to exit code=1

Spartakovic (approved):

Issue #60, opened 2023-03-23
Support digger.yaml along with digger.yml 


Spartakovic:
We currently read a configuration file for specifying config options. This configuration file is specified directly in the repository. We only support digger.yml at the moment, we want to also support digger.yaml along with it if it exists.

If both files exists then its good to throw an error saying that you are not sure which one to chose, please delete one of the files.

PR #66, opened 2023-03-25
yaml and yml both are accepted as config file #60

ondbyte:
it's a error condition if both yaml and yml exists


motatoes:
Thank you for your contribution! Unfortunately this has already been implemented in https://github.com/diggerhq/digger/pull/67

PR #67, opened 2023-03-26
Support both digger.[yml | yaml]

### LINKED ISSUE #60: Support digger.yaml along with digger.yml

ynden:
### Changelog
Fixes #60 

#### Fix details
* Added support for both digger.yaml & digger.yml config files
* Added digger config file conflict detection
* Added a specific error for  when multiple DiggerConfig are detected
* Added tests to cover multiple digger config files case
* Added tests to validate `digger.yaml` is now recognized as a legit config file
* Fixed a test message error

#### Refactor
* Removed nested conditions from `NewDiggerConfig` function
* Removed else conditions from `NewDiggerConfig` function
* Created SetUp & TearDown test functions
* Removed path concatenation, and used `path.Join` instead

### DIFF
--- a/pkg/digger/digger_config_test.go
+++ b/pkg/digger/digger_config_test.go
@@ -3,10 +3,18 @@ package digger
 import (
 	"log"
 	"os"
+	"path"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
+
+func setUp() (string, func()) {
+	tempDir := createTempDir()
+	return tempDir, func() {
+		deleteTempDir(tempDir)
+	}
+}
 
 func TestDiggerConfigFileDoesNotExist(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/digger/digger_config_test.go
```diff
@@ -48,7 +92,7 @@ projects:
 	}
 
 	dg, err := NewDiggerConfig(tempDir)
-	assert.NoError(t, err, "expected error to be not nil")
+	assert.NoError(t, err, "expected error to be nil")
```
ynden, line 103: I think there was a typo here

PR #70, opened 2023-03-28
Return full path & refactor file creation in tests

ynden:
#### Fix
* Return full path to config instead of just the name

#### Refactor
* Added `createFile` function to create & delete easily while testing


veziak (approved):
lgtm

PR #76, opened 2023-03-28
support terragrunt

Spartakovic:


pkg/terraform/tf.go
```diff
@@ -1,24 +1,55 @@
...
+	return terragrunt.runTerragruntCommand("apply")
+}
+
+func (terragrunt Terragrunt) Plan() (bool, string, string, error) {
+	stdout, stderr, err := terragrunt.runTerragruntCommand("plan")
+	return true, stdout, stderr, err
+}
+
+func (terragrunt Terragrunt) runTerragruntCommand(command string) (string, string, error) {
+	cmd := exec.Command("terragrunt", command, "--terragrunt-working-dir", terragrunt.WorkingDir)
```
motatoes, line 37: is terragrunt executable available in the github action by default?

Spartakovic: nope, I assumed users will install it with some other action

Spartakovic: should we install it?

motatoes: You're right, I think it makes sense to install prior step, lets merge it and test it out

motatoes (approved):

PR #77, opened 2023-03-28
fix happy path test

veziak:


pkg/integration/integration_test.go
```diff
@@ -410,7 +392,7 @@ func TestHappyPath(t *testing.T) {
 	resource := repositoryName + "#default"
```
Spartakovic: this should be +#dev given projectLock config above

veziak: updated pr

Spartakovic (approved):

PR #91, opened 2023-03-31
feat: support workflows and extra args to terraform executions

Spartakovic:


pkg/digger/digger_config.go
```diff
@@ -18,15 +17,30 @@ type WorkflowConfiguration struct {
...
+	Steps []Step `yaml:"steps"`
+}
+
+type Workflow struct {
+	Plan          *Stage                 `yaml:"plan,omitempty"`
+	Apply         *Stage                 `yaml:"apply,omitempty"`
+	Configuration *WorkflowConfiguration `yaml:"workflow_configuration"`
+}
+
+type Command struct {
```
motatoes: this isn't used anywhere, its part of a step now, should it be removed?

pkg/digger/digger_config.go
```diff
@@ -50,6 +56,28 @@ func (p *Project) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 }
 
+type Step struct {
+	Action    string
```
motatoes, line 60: this needs `yaml:"action"` appended

Spartakovic: no, it's not 1-2-1 mapping between yaml and go struct

veziak (approved):
lgtm

Issue #18, opened 2023-03-03
implement digger help command


motatoes:
show what are the possible commands to use when user comments `digger` or `digger help` or another incorrect digger command

veziak:
"help" and "version" for a start

motatoes:
digger help

```
Use the following commands to get started:

digger version: Show digger version
digger plan:
digger apply: 
digger unlock: 
```

digger version

```
Using digger version 0.2.1
```

Version can be hardcoded as a constant for now and we will bump it for every release

PR #96, opened 2023-04-01
Implement Help and Version Commands 

s1ntaxe770r:
This PR adds version and help commands to the digger CLI
As requested in  #18

### DIFF
--- a/cmd/digger/main.go
+++ b/cmd/digger/main.go
@@ -11,6 +11,16 @@ import (
 )
 
 func main() {
+	args := os.Args[1:]
+	if len(args) > 0 && args[0] == "version" {
+		fmt.Println(utils.GetVersion())
+		os.Exit(0)
+	}
+	if len(args) > 0 && args[0] == "help" {
+		utils.DisplayCommands()
+		os.Exit(0)
+	}
+
 	diggerConfig, err := digger.NewDiggerConfig("")
 	if err != nil {
 		fmt.Printf("Failed to read digger config. %s\n", err)
--- a/cmd/digger/main.go
+++ b/cmd/digger/main.go
@@ -67,10 +67,9 @@ func main() {
 	}
 	println("GitHub event processed successfully")
 
-	if eventName == "pull_request" && parsedGhContext.Action == "commented" {
+	if digger.CheckIfHelpComment(ghEvent) {
 		reply := utils.GetCommands()
-		githubPrService.ReplyComment(prNumber, reply)
-		os.Exit(0)
+		githubPrService.PublishComment(prNumber, reply)
 	}
 
 	commandsToRunPerProject, err := digger.ConvertGithubEventToCommands(ghEvent, impactedProjects)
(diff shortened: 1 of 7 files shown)


pkg/utils/commands.go
```diff
@@ -0,0 +1,23 @@
...
+	Name        string
+	Description string
+}
+
+var availableCommands = []Commad{
+	{"digger help", "Display help information"},
+	{"digger version", "Display version information"},
+	{"digger apply", "Apply digger configuration"},
+	{"digger plan", "Plan digger configuration"},
+	{"digger lock", "Lock Terraform state"},
```
Spartakovic: there is digger unlock too. Also it doesn't lock terraform state, it rather locks terraform projects on pull request level

pkg/utils/commands.go
```diff
@@ -0,0 +1,23 @@
...
+type Commad struct {
+	Name        string
+	Description string
+}
+
+var availableCommands = []Commad{
+	{"digger help", "Display help information"},
+	{"digger version", "Display version information"},
+	{"digger apply", "Apply digger configuration"},
+	{"digger plan", "Plan digger configuration"},
```
Spartakovic: it doesn't apply digger configuration, but terraform configuration. It just adds few more things on top, like pull request locking

motatoes:
In addition to being invoked as a cli command, digger is also invoked as part of github action in which case the arguments are github context. We also want to react to a comment in the pull request and publish a comment in result. We can do this [here](https://github.com/diggerhq/digger/blob/3fffbdcc2073de471f92e2f3a634b251008cbdd0/cmd/digger/main.go#L59-L58) after detecting that the event is a comment and the comment is "digger help" then publish a comment as a result to it and exit(0)

cmd/digger/main.go
```diff
@@ -57,6 +67,12 @@ func main() {
 	}
 	println("GitHub event processed successfully")
 
+	if eventName == "pull_request" && parsedGhContext.Action == "commented" {
+		reply := utils.GetCommands()
+		githubPrService.ReplyComment(prNumber, reply)
+		os.Exit(0)
+	}
```
motatoes: I suggest to do it like this

```suggestion
	if digger.CheckIfHelpComment(ghEvent) {
		reply := utils.GetCommands()
		githubPrService.PublishComment(prNumber, reply)
	}
```

Where you can define the function of `CheckIfHelpComment` as follows in digger.go

```
func CheckIfHelpComment(event models.Event) bool {
	switch event.(type) {
	case models.IssueCommentEvent:
		event := event.(models.IssueCommentEvent)
		if strings.Contains(event.Comment.Body, "digger help") {
			return true
		}
	}
	return false
}
```

pkg/github/github.go
```diff
@@ -24,6 +25,7 @@ type GithubPullRequestService struct {
 type PullRequestManager interface {
 	GetChangedFiles(prNumber int) ([]string, error)
 	PublishComment(prNumber int, comment string)
+	ReplyComment(prNumber int, reply string)
```
motatoes: This function is not needed at all

motatoes (changes requested):
few last changes, but looking good so far

motatoes (approved):

Issue #75, opened 2023-03-28
Support more lock providers


ynden:
`digger` currently supports  DynamoDB for locks. However, in Azure DynamoDB service doesn't exist (there's a CosmosDB, but not the same schema).

A more native way in Azure would be to use a Storage Account. Actually the same storage account where the terraform state is kept. We would rely on Storage Account table to store our lock.

But before that, we need to adjust the codebase to start using multiple cloud providers.

* Start using a LockService struct
* Add CloudLock interface as property

Then instead of using DynamoDB directly, we'll call LockService which will rely on the underlying CloudLock provider.
The right lock provider will be injected when creating a LockService.

Example of implementation, check below
```go
// Add the right param and return type
type CloudLock interface {
	Lock()
        Unlock()
        ForceUnlock()
}
```
The Lock Service:

```go
type LockService struct {
	lockProvider CloudLock
}

func (ls *LockService) Lock() {
    return ls.lockProvider.Lock()
```

StorageAccount in Azure, implementation example:
```go
type StorageAccountLock struct {
}

func (sal *StorageAccountLock) Lock()  {
	// add table with data to storage account
}
```

Full workflow (kept easy to be readable):

```go
//Detect cloud vendor (or will be defined in digger.yaml)
lockVendor := getLockVendor(&config)
var cloudLock CloudLock

// Use switch or if/else
if lockVendor == "azure" {
    cloudLock := cloudLock.(StorageAccountLock)
}

if lockVendor == "aws" {
     cloudLock := cloudLock.(DynamoDBLock)
}

lockService := LockService {
    lockProvider: cloudLock
}
lockService.Lock()
```

Spartakovic:
There is already a Lock interface pkg/utils/locking

motatoes:
Thank you for this detailed issue and outlining refactoring steps @ynden. All looks good here, looking forward to reading your PR

motatoes:
performed in a separate branch and will be merged gradually into mainline as we restructure bits and peices

Issue #74, opened 2023-03-28
Support more SCM vendors


ynden:
`digger` currently support Github, but we also want to extend it to other SCM vendors:
* Bitbucket
* Azure DevOps
* GitLab

However, to do that the codebase needs to be adapted:
* Start using a PullRequestService struct
* Add SCMProvider interface as property

Then instead of using Github or Bitbucket directly, we'll call PullRequestService which will rely on the underlying SCM provider.
The right SCM provider will be injected when creating a PullRequestService.

For example, check below:

There seems to already be a PullRequestManager (our interface):
```go
type PullRequestManager interface {
	GetChangedFiles(prNumber int) ([]string, error)
	PublishComment(prNumber int, comment string) error
}
```

The PR Service:
```go
type PullRequestService struct {
	pr PullRequestManager
}

func (prs *PullRequestService) PublishComment(prNumber int, comment string) error {
    return prs.pr.PublishComment(prNumber, comment)
```

Bitbucket implementation example:
```go
type Bitbucket struct {
}

func (bb *Bitbucket) GetChangedFiles(prNumber int) ([]string, error) {
	return []string{}, nil
}

func (bb *Bitbucket) PublishComment(prNumber int, comment string) error {
	return nil
}
```

Github implementation example:
```go
type Github struct {
}

func (gh *Github) GetChangedFiles(prNumber int) ([]string, error) {
	return []string{}, nil
}

func (gh *Github) PublishComment(prNumber int, comment string) error {
	return nil
}
```

Full workflow (kept easy to be readable):
```go
//Detect scm vendor (or will be defined in digger.yaml)
scmVendor := getScmVendor(&config)
var prManager PullRequestManager

// Use switch or if/else
if scmVendor == "Bitbucket" {
    prManager := prManager.(Bitbucket)
}

if scmVendor == "Github" {
     prManager := prManager.(Github)
}

prService := PullRequestService {
    pr: prManager
}
prService.PublishComment(2, "digger apply")
```

veziak:
sounds like a good plan!

veziak:
just a quick note that there is another bit of functionality related to SCMs: events data that digger read from env vars, GitHub Context for example. I'm implementing similar thing for Bitbucket Contexts at the moment(for the case when we trigger Bitbucket pipeline using webhooks)

motatoes:
Thank you for this detailed issue and outlining refactoring steps @ynden. All looks good here, looking forward to reading your PR

[...]

PR #98, opened 2023-04-03
Refactor: Restructure core components

ynden:
### Changelog
* Rework digger core
* Separate components into what they do (ci, scm, runner, etc.)
* Add dependency injection mecanism
* Reduce coupling between components
* Decouple tools from business use cases

Because of #74 & #75, we wanted to refactor the code to make it extensible, and the condition was to add services instead of using concrete implementations. However, before even doing that, we need to restructure `digger` and its core components, because components are tightly coupled.

The current work is still in progress, but comes to re-structure the core of `digger` (even though what have been implemented before works well, it's only to allow implementing new features more quicker).

With the new structure, it'll become easier to extend what digger can do, because of the core components being now loosely coupled.

I recommend branching out the current branch to better understand the changes.


pkg/domain/scm_provider.go
```diff
@@ -0,0 +1,5 @@
+package domain
+
+type SCMProvider interface {
```
veziak, line 3: Should we call it CIService instead of SCMProvider? I think it's better reflecting what it is. BitBucket, GitLab, GitHub are more CI tools than SCM in our case? And implementations would be BitBucketProvider, GitLabProvider, GitHubProvider in this case. Not sure myself what's the best naming, open for discussion.

motatoes: I prefer CI as a name too. Github and Bitbucket are SCM tools but *Github Actions* *Bitbucket Pipelines* *Jenkins* are CI tools and digger runs in CI systems hence CI is a better name in this case

ynden: In this implementation, there is a package called `ci_runner` just for what you just mentioned. There is also the package `SCMProvider`, that handles the pull request work, publish a comment for example.

CI runner on the other side handles the event received (from the SCMProvider)

So you can have your SCM as Github, but your CI runner as Jenkins. Only Jenkins knows what kind of environment variables are there, specific to Jenkins. So we'll be handling events on Jenkins, but publishing comments on Github, hence the differentiation.

I don't know if that makes sense, or should we change it?

cmd/digger/main.go
```diff
@@ -1,73 +1,54 @@
...
-
-	ghContext := os.Getenv("GITHUB_CONTEXT")
-	if ghContext == "" {
-		fmt.Printf("GITHUB_CONTEXT is not defined. %s\n", err)
-		os.Exit(4)
-	}
-
-	parsedGhContext, err := models.GetGitHubContext(ghContext)
+	// Parse event
+	ps := service.NewParser()
```
Spartakovic: seems like parser is disconnected from CI runner, although I think it's coupled by nature. Maybe it's ci runner that needs to return appropriate parser, or maybe it just needs to return parsed event. 

I cannot imagine that we will parse github event in 2 different ways for example.

cmd/digger/main.go
```diff
@@ -1,73 +1,54 @@
...
 	if err != nil {
-		fmt.Printf("failed to convert event to command, %v", err)
-		os.Exit(7)
+		log.Fatalf("got an error while getting the Lock provider: %v", err)
 	}
-	println("GitHub event converted to commands successfully")
 
-	err = digger.RunCommandsPerProject(commandsToRunPerProject, repoOwner, repositoryName, eventName, prNumber, diggerConfig, githubPrService, lock, "")
+	// Inject the dependencies
+	prs := service.NewPullRequest().
```
Spartakovic: it seems like everything is revolving around pull requests. Although it's true right now, the real goal of the project is to run terraform, and things like pull request locks will be a like an optional plugin. Same with leaving comments on pull requests. The main part is running terraform, the rest is optional (eventually)

pkg/tf_runner/terraform.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (terraform *Terraform) SetWorkingDir(path string) error {
+	if path == "" {
+		return fmt.Errorf("path is empty")
+	}
+	terraform.WorkingDir = path
+	return nil
+}
+
+func (terraform *Terraform) Plan() (*domain.TerraformOutput, error) {
```
Spartakovic: terraform runner will need to accept extra parameters soon, so will need to incorporate this into your design

cmd/digger/main.go
ynden: Absolutely! I agree on that, the CI Runner should be returning the parsed event. Thanks for the suggestion

cmd/digger/main.go
ynden: I did think that the features were around pull requests. In that case I'll change the current behaviour to reflect your suggestions in the code. Thanks

motatoes:
I am closing this since it is now pushed to the branch `feature/restructure-core` on this repository and we will be gradually taking some changes from that branch and merging it into our repository in terms of the new interface structure

PR #104, opened 2023-04-05
[ISS-92] Add a flag to control usage data collection 

### LINKED ISSUE #92: Add option for disabling usage data
We want to add options for disabling usage data collection using either environment variable or an option in `digger.yml` config [pkg/utils/usage.go](https://github.com/diggerhq/digger/blob/main/pkg/utils/usage.go).

Environment variable

```
COLLECT_USAGE_DATA=true # default value
```

the above will set the global setting for 

digger.yml config

```
collect_usage_data: true
projects:
- name: dev
   ....

```

setting it in the config will override any environment variable that was previously set. The value will be true by default.

Spartakovic:
This pull request adds functionality to control the collection of user data usage via either an environment variable or a config file field. Users can now set the COLLECT_USAGE_DATA variable to "true" or "false" or update the corresponding field in the config file to enable or disable data usage collection.

closes #92


veziak (approved):
lgtm

PR #106, opened 2023-04-05
[ISS-94] require GOOGLE_STORAGE_BUCKET env var to be set if using gcp

### LINKED ISSUE #94: Get rid of hardcoded value of bucket for GCP
GCP bucket names are globally unique. We assumed that they would not be globally unique. We had hardcoded the value of bucket as `digger-lock` so now we want to instead make the user define it using the environment name directly

if the environment variable is not set and we are using GCP then we throw an error asking the user to set this value GOOGLE_STORAGE_BUCKET

Spartakovic:
We used to default to digger-lock as a bucket name, however gcp bucket names are supposed to be unique

closes #94


veziak (approved):
lgtm

Issue #111, opened 2023-04-06
Testing Terragrunt


daanbosch:
I keep getting the following error:
```
Error locking project: ResourceNotFoundException: Requested resource not found
```

Any way to debug what is going on?

```
projects:
- name: 00_myvm
  dir: dev/env/00_myvm
  terragrunt: true
```

ZIJ:
hey @daanbosch thanks for filing the issue! did you add your cloud provider keys as secrets in the action? (assuming you are using it with Github Actions)

also - what other log outputs are there?

daanbosch:
Full logs:
```
Run diggerhq/tfrun@main
Run cd /runner/_work/_actions/diggerhq/tfrun/main
go: downloading github.com/aws/aws-sdk-go v1.44.219
go: downloading cloud.google.com/go/storage v1.30.1
go: downloading github.com/google/go-github/v50 v50.1.0
go: downloading gopkg.in/yaml.v2 v2.4.0
go: downloading github.com/hashicorp/terraform-exec v0.18.1
go: downloading github.com/hashicorp/terraform-json v0.15.0
go: downloading github.com/hashicorp/go-version v1.6.0
go: downloading cloud.google.com/go v0.110.0
go: downloading golang.org/x/oauth2 v0.6.0
go: downloading golang.org/x/crypto v0.5.0
go: downloading github.com/google/go-querystring v1.1.0
go: downloading github.com/zclconf/go-cty v1.13.0
go: downloading cloud.google.com/go/iam v0.12.0
go: downloading github.com/google/uuid v1.3.0
go: downloading github.com/googleapis/gax-go/v2 v2.7.1
go: downloading google.golang.org/api v0.114.0
go: downloading cloud.google.com/go/compute/metadata v0.2.3
go: downloading cloud.google.com/go/compute v1.18.0
go: downloading google.golang.org/genproto v0.0.0-20230320184635-7606e756e683
go: downloading google.golang.org/grpc v1.53.0
go: downloading google.golang.org/protobuf v1.29.1
go: downloading go.opencensus.io v0.24.0
go: downloading golang.org/x/xerrors v0.0.0-20220907171357-04be3eba64a2
go: downloading golang.org/x/net v0.8.0
go: downloading github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e
go: downloading github.com/apparentlymart/go-textseg/v13 v13.0.0
go: downloading golang.org/x/text v0.8.0
[...]

PR #116, opened 2023-04-06
create dynamodb table if it does not exists

motatoes:
DynamoDB table is not being created during the first run as mentioned in #111 resulting in `ResourceNotFound` Exception. This Pull request creates the table during any of the locking operations if it does not yet exist already. 

TODO: 

- Make the table creation only the first time (maybe using a flag of some sort?) Since otherwise there will be too much AWS calls happening for every action to check if the table exists.
- Write additional tests for this 
- Ensure there is no race conditions for the first run since this often occurs due to eventual consistency of AWS

### DIFF
--- a/pkg/aws/dynamo_locking.go
+++ b/pkg/aws/dynamo_locking.go
@@ -1,9 +1,11 @@
 package aws
 
 import (
+	"fmt"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/aws/aws-sdk-go/service/dynamodb/expression"
+	"os"
 	"strconv"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


pkg/aws/dynamo_locking.go
```diff
@@ -14,7 +16,39 @@ type DynamoDbLock struct {
 	DynamoDb *dynamodb.DynamoDB
 }
 
+func (dynamoDbLock *DynamoDbLock) createTableIfNotExists() {
+	input := &dynamodb.CreateTableInput{
```
veziak: would it be better if we check if table exist explicitly?

veziak: would be good to log something if table has been created

pkg/aws/dynamo_locking.go
```diff
@@ -14,7 +16,39 @@ type DynamoDbLock struct {
 	DynamoDb *dynamodb.DynamoDB
 }
 
+func (dynamoDbLock *DynamoDbLock) createTableIfNotExists() {
```
Spartakovic, line 19: I think it's misleading to call it createTableIfNotExists() given it doesn't check that, might as well just call it createTable()

veziak (approved):
lgtm

PR #147, opened 2023-04-19
add generate_projects option to digger.yaml

veziak:
```
projects:
- name: dev
- 
...

generate_projects:
  include: us-east1
  exclude: us-east1/digger-test1
```

```
us-east-1
      modules/
      digger-test1/
           main.tf
           config.tf
      digger-test2/
           main.tf
          config.tf
    
```


pkg/digger/digger_config.go
```diff
@@ -46,6 +53,40 @@ type Workflow struct {
 	Configuration *WorkflowConfiguration `yaml:"workflow_configuration"`
 }
 
+type DirWalker interface {
+	GetDirs(workingDir string) ([]string, error)
+}
+
+type MockDirWalker struct {
```
Spartakovic: this should live in _test file probably

pkg/digger/digger_config.go
```diff
@@ -46,6 +53,40 @@ type Workflow struct {
...
+	Files []string
+}
+
+type FileSystemDirWalker struct {
+}
+
+func (walker *FileSystemDirWalker) GetDirs(workingDir string) ([]string, error) {
+	var files []string
+	err := filepath.Walk(workingDir,
```
Spartakovic, line 65: do we need to walk whole tree? or just first level? If they have subfolders in subfolders they will be treated as different projects. Maybe we should only support one level

veziak: I think we need to walk the whole tree, otherwise 'include' pattern will be limited only to one dir name (first level) ?

pkg/digger/digger_config.go
veziak: yep, just moved it

pkg/digger/digger_config.go
Spartakovic: yeah that's my question. Should we only support one level? Because if people will structure their infra in folders we will start deploying each one as a separate project which I think is a bad idea

Spartakovic (approved):

PR #153, opened 2023-04-20
fix example workflow config

Spartakovic:
generally follows the same structure as atlantis

https://www.runatlantis.io/docs/custom-workflows.html#adding-extra-arguments-to-terraform-commands


veziak (approved):
lgtm

PR #169, opened 2023-04-25
feat: post collapsible plan and apply output PR comments

fleroux514:
The output of `terraform plan` and `terraform apply` commands can be very big, and could clutter the PR everytime being posted in comments. Therefore making use of collapsible comments instead:

image

image

### DIFF
--- a/pkg/digger/digger.go
+++ b/pkg/digger/digger.go
@@ -370,7 +370,7 @@ func (d DiggerExecutor) Plan(prNumber int) error {
 					return fmt.Errorf("error executing plan: %v", err)
 				}
 				plan := cleanupTerraformPlan(isNonEmptyPlan, err, stdout, stderr)
-				comment := "Plan for **" + d.lock.LockId() + "**\n" + plan
+				comment := utils.GetCollapsibleComment("Plan for **"+d.lock.LockId()+"**", plan)
 				d.prManager.PublishComment(prNumber, comment)
 			}
 			if step.Action == "run" {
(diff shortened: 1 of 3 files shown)


pkg/digger/digger.go
```diff
@@ -518,8 +518,7 @@ func cleanupTerraformOutput(nonEmptyOutput bool, planError error, stdout string,
 		endPos = strings.Index(stdout, matches[0]) + len(matches[0])
 	}
 
-	result = stdout[startPos:endPos]
-	return "```terraform\n" + result + "\n```"
```
fleroux514, line 520: Not sure the impact of removing this "wrapper" formatting or if its still needed with the current change.

With it I would get 

image

pkg/utils/comments.go
```diff
@@ -0,0 +1,13 @@
+package utils
+
+func GetCollapsibleComment(summary string, collapedComment string) string {
```
Spartakovic: typo

pkg/utils/comments.go
```diff
@@ -0,0 +1,13 @@
...
+	str := `<details>
+  <summary>` + summary + `</summary>
+  
+  ` + "```" + `
```
Spartakovic: as far as understand you can use ```terraform here

basically collapsible input still support code blocks in it

Spartakovic: and we can rename it to GetTerraformOutputAsCollapsibleComment, or make it generic and pass a flag

pkg/utils/comments.go
```diff
@@ -1,11 +1,11 @@
 package utils
 
-func GetCollapsibleComment(summary string, collapedComment string) string {
+func GetTerraformOutputAsCollapsibleComment(summary string, collapsedComment string) string {
 	str := `<details>
   <summary>` + summary + `</summary>
   
-  ` + "```" + `
-` + collapedComment + `
+  ` + "```terraform" + `
```
fleroux514, line 7: adding it here seems to break the collapsible functionality

image

veziak:
just tested it manually, if comment looks like this :

image

then it works as expected, it's important that there is a new line between summary and terraform line:

<details>
  <summary>Plan</summary>

```terraform
resource "null_resource" "test" {
}
```
</details>

fleroux514:
> just tested it manually, if comment looks like this :
> 
> image
> 
> then it works as expected, it's important that there is a new line between summary and terraform line:
> 
> Plan

Thanks @veziak, may have been a testing error on my part, it does work.

pkg/utils/comments.go
fleroux514: fixed

pkg/utils/comments.go
fleroux514: done

Spartakovic (approved):

PR #172, opened 2023-04-26
fix: plan filename search

Spartakovic:
When github action uploads files it includes relative path.
To match that we can just match the name of the file instead of the whole path, given that project names are unique. Which they should be


veziak (approved):
lgtm

PR #183, opened 2023-04-29
fix: adding logic to validate if the PR must get merged

fleroux514:
Added validation:
- check if apply comment along with auto_merge config before trying to merge the PR
- check if a requested project name was specified in the comment, and if the PR impacts more than one project (in which case we don't want to merge)


veziak (approved):
lgtm

PR #184, opened 2023-05-01
fix: change regex used to parse project names

fleroux514:
Using a generic regex to match anything after a `digger plan/apply -p` comment, instead of specific characters. Or else permitted characters for projects names declared in digger.yml need to be documented and the project names need to be validated using the same regex.

Fixes the current issue
image


veziak:
@fleroux514 let's add just underscore for a start? I'm not sure what side effects we are going to have if it's too broad. We probably don't want users to use spaces, quotes etc in a project name

PR #199, opened 2023-05-03
fix: disable project unlock on apply default functionality

### LINKED ISSUE #185: Possibility to unlock project on PR merge instead of apply command
Right now a project is unlocked as soon as the apply cmd succeeded.

I'm thinking this might be problematic when the PR affects more than one project and
- user applies changes on one specific project  (`digger apply -p xxx`)
- user runs `digger apply` but some projects fail and some succeed.

This might leave some projects with a temporary drift since their state file is not in sync with the code, and allows new PRs to grab the lock and possibly get merged  before the original PR is merged into main.

fleroux514:
This behavior does not exist with Atlantis, so removing it accordingly,

In digger.yml, `on_commit_to_default` can be configured to release locks when the PR is merged instead.
```
    workflow_configuration:
      on_commit_to_default: [digger unlock]

```

closes #185

### DIFF
--- a/pkg/digger/digger.go
+++ b/pkg/digger/digger.go
@@ -445,13 +445,15 @@ func (d DiggerExecutor) Apply(prNumber int) error {
 					applyOutput := cleanupTerraformApply(true, err, stdout, stderr)
 					comment := utils.GetTerraformOutputAsCollapsibleComment("Apply for **"+d.lock.LockId()+"**", applyOutput)
 					d.prManager.PublishComment(prNumber, comment)
-					if err == nil {
-						_, err := d.lock.Unlock(prNumber)
-						if err != nil {
-							return fmt.Errorf("error unlocking project: %v", err)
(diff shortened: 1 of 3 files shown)


pkg/digger/digger.go
```diff
@@ -445,14 +445,6 @@ func (d DiggerExecutor) Apply(prNumber int) error {
 					applyOutput := cleanupTerraformApply(true, err, stdout, stderr)
 					comment := utils.GetTerraformOutputAsCollapsibleComment("Apply for **"+d.lock.LockId()+"**", applyOutput)
 					d.prManager.PublishComment(prNumber, comment)
-					if err == nil {
-						_, err := d.lock.Unlock(prNumber)
-						if err != nil {
-							return fmt.Errorf("error unlocking project: %v", err)
-						}
-					} else {
```
Spartakovic, line 453: we still want to publish comment about failed apply though, just not unlock it

fleroux514: Re-added

PR #209, opened 2023-05-05
feat: add env variable support

Spartakovic:


pkg/configuration/digger_config.go
```diff
@@ -50,11 +56,17 @@ type Stage struct {
 }
 
 type Workflow struct {
+	Envs          Envs                   `yaml:"envs"`
```
veziak: Should we call it EnvVars? just to avoid confusion with any kind of "environments"

veziak (approved):
lgtm

PR #216, opened 2023-05-08
fix: prevent error about lock being acquired by the current PR

fleroux514:
Prevent the following

image

### DIFF
--- a/pkg/gcp/gcp_lock.go
+++ b/pkg/gcp/gcp_lock.go
@@ -1,12 +1,13 @@
 package gcp
 
 import (
-	"cloud.google.com/go/storage"
 	"context"
 	"fmt"
 	"log"
 	"strconv"
 	"time"
+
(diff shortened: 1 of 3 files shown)


pkg/gcp/gcp_lock.go
```diff
@@ -20,10 +21,14 @@ func (googleLock *GoogleStorageLock) Lock(transactionId int, resource string) (b
 	if err != nil {
 		fmt.Printf("failed to get lock: %v\n", err)
 	}
-	if existingLockTransactionId != nil {
+	if existingLockTransactionId != nil && *existingLockTransactionId != transactionId {
```
Spartakovic: I think this kind of logic should actually live a layer above, otherwise we will have to repeat it for each lock provider

fleroux514: Right, fixed

Spartakovic (approved):

PR #248, opened 2023-05-12
fix: auto-merge logic when apply is executed with no specific project

fleroux514:
digger would not auto-merge the PR when commenting `digger apply` without specifying a project


Spartakovic:
I am questioning the whole mergePrIfCmdSuccessfull, I don't quite understand it's original purpose

Spartakovic (approved):

PR #250, opened 2023-05-12
fix: post pr comment when a custom workflow command fails in plan or apply stage

fleroux514:
image


fleroux514:
hmmm, just retested this and noticed there was another PR comment. will close the current pr

image

fleroux514:
Reopening this PR since the current comment does not specify which command failed and sometimes only returns the error msg..

image

PR #251, opened 2023-05-12
feat: support choosing shell for run scripts

Spartakovic:
Support choosing shell when running custom commands
```
- run: source venv/bin/activate
  shell: bash
```


fleroux514 (approved):
LGTM!

PR #252, opened 2023-05-12
change logic around deciding when to merge PR

Spartakovic:


pkg/digger/digger.go
```diff
@@ -247,11 +248,16 @@ func ConvertGithubEventToCommands(event models.Event, impactedProjects []configu
 				})
 			}
 		}
-		return commandsPerProject, nil
+		return commandsPerProject, true, nil
 	case models.IssueCommentEvent:
 		event := event.(models.IssueCommentEvent)
 		supportedCommands := []string{"digger plan", "digger apply", "digger unlock", "digger lock"}
 
+		if requestedProject != nil {
```
fleroux514, line 262: If think `&& requestedProject.Name == impactedProjects[0].Name` is automatically implied if `len(impactedProjects) == 1`

Instead maybe log an error if ` requestedProject.Name != impactedProjects[0].Name` which would mean there is a bug..

Spartakovic: I wonder if failing fast is still better

Spartakovic: defensive programming :D

fleroux514: I understand. There probability of an issue related to this is low.  Let's keep it as is.

fleroux514 (approved):
LGTM

veziak (approved):
lgtm

PR #255, opened 2023-05-15
fix build

veziak:


pkg/digger/digger.go
```diff
@@ -46,7 +46,7 @@ func ProcessGitHubEvent(ghEvent models.Event, diggerConfig *configuration.Digger
 			}
 			impactedProjects = newImpactedProjects
 		} else {
-			mergePrIfCmdSuccessfull = true
```
motatoes, line 49: maybe just rename to fix typo instead `mergePrIfCmdSuccessful` with single L

motatoes (approved):

PR #277, opened 2023-05-19
fix #260 workflow is misconfigured

### LINKED ISSUE #260: Fail with clearer error message when digger.yml workflow is misconfigured
Consider this digger.yml

```
projects:
  - name: dev
    dir: ./dev
    workflow: devworkflow

workflows:
  devworkflow:
    plan:
      steps:
      - run: "mv dev/dev.tfvars dev/dev.auto.tfvars"
      - init:
      - plan:
```

This is not a valid workflow since `init:` and `plan:` should be applied without any trailing columns, currently this fails in v18 with obscure error:

```
Error: Backend initialization required, please run "terraform init"
```

Better would be to fail earlier and say something like:

```
Invalid configuration, missing init found in `defaworkflow`, please ensure that you specify both init and plan in your configuration file
```

Rouphen:
1. add Unit Test TestNonInitConfiguration
2. add step checking 
3. add DiggerConfig-checking
4. call the DiggerConfig-checking when new a DiggerConfig
5. add s.extract(stepMap, "init")


Rouphen:
linked issue #260

pkg/configuration/digger_config.go
```diff
@@ -388,6 +407,41 @@ func (c *DiggerConfig) GetWorkflowConfiguration(projectName string) WorkflowConf
...
+	return nil
+}
+
+func (c *DiggerConfig) CheckWorkflowConfiguration() error {
+	for _, project := range c.Projects {
+		workflows := c.Workflows
+		workflow, ok := workflows[project.Workflow]
+
+		if !ok {
+			return nil
```
motatoes, line 428: here we should raise an error of "project specifies workflow not in list of workflows

pkg/configuration/digger_config.go
```diff
@@ -388,6 +407,41 @@ func (c *DiggerConfig) GetWorkflowConfiguration(projectName string) WorkflowConf
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (c *DiggerConfig) CheckWorkflowConfiguration() error {
+	for _, project := range c.Projects {
+		workflows := c.Workflows
```
motatoes, line 424: This line can move out of the for loop

pkg/configuration/digger_config.go
```diff
@@ -388,6 +407,41 @@ func (c *DiggerConfig) GetWorkflowConfiguration(projectName string) WorkflowConf
 	return *workflow.Configuration
 }
 
+func checkSteps(steps []Step) error {
```
motatoes, line 410: I think that the checks in this function are incorrect because it should be checking that for a plan flow: at least `init` and `plan` exists

for apply it should check that at least `init` and `apply` flow exists

motatoes (changes requested):
please take a look at the comments that I have added

pkg/configuration/digger_config.go
Rouphen: do you means, the plan and apply need difference checking, e.g.

```
func checkPlanSteps(steps []Step) error {
	for _, step := range steps {
		if (step.Action == "init" || step.Action == "plan") && (step.Value == "" && step.ExtraArgs == nil) {
			return fmt.Errorf("%s", step.Action)
		}
	}

	return nil
}

func checkApplySteps(steps []Step) error {
	for _, step := range steps {
		if (step.Action == "init" || step.Action == "apply") && (step.Value == "" && step.ExtraArgs == nil) {
			return fmt.Errorf("%s", step.Action)
		}
	}

	return nil
}

func (c *DiggerConfig) CheckWorkflowConfiguration() error {
	workflows := c.Workflows

	for _, project := range c.Projects {
		workflow, ok := workflows[project.Workflow]

		if !ok {
			return fmt.Errorf("project '%s' specifies workflow not in list of workflows", project.Name)
		}

		err := checkPlanSteps(workflow.Plan.Steps)
		if err != nil {
			return fmt.Errorf(ErrDiggerInvalidWorkflow.Error(), project.Workflow, err.Error())
		}

		err = checkApplySteps(workflow.Apply.Steps)
		if err != nil {
			return fmt.Errorf(ErrDiggerInvalidWorkflow.Error(), project.Workflow, err.Error())
		}
	}

	return nil
}
```

motatoes: No its more like that for every workflow we need to check that there is BOTH init and plan. In your function the check is only for a single check. The logic is like this:

```
for all workflows {
     check if there is a plan step then {
            check if both init and plan steps exist
            if they don't exist then throw error
     }
    check if there is apply step then {
         if apply step exists then both init and apply steps exist
    }
}
```

Hope that helps

Rouphen: thanks, I got.

and I noticed, there are some unit tests, e.g. "TestDiggerConfigWhenOnlyYamlExists", is it invalid when we apply new checking? 
func TestDiggerConfigWhenOnlyYamlExists(t *testing.T) {
	tempDir, teardown := setUp()
	defer teardown()

	diggerCfg := `
projects:
- name: prod
  branch: /main/
  dir: path/to/module/test
  workspace: default
`
	deleteFile := createFile(path.Join(tempDir, "digger.yaml"), diggerCfg)
	defer deleteFile()

	dg, err := NewDiggerConfig(tempDir, &FileSystemDirWalker{})
	assert.NoError(t, err, "expected error to be nil")
	assert.NotNil(t, dg, "expected digger config to be not nil")
	assert.Equal(t, "path/to/module/test", dg.GetDirectory("prod"))
}

Issue #99, opened 2023-04-04
Making supply of Keys optional by making locking optional


motatoes:
Currently the requirement to support cloud account keys (AWS, GCP, Azure) is due to the fact that we need them to provision and use resources for locking. This adds extra cumbersome steps of getting keys and supplying them after forking the [demo repository](https://github.com/diggerhq/digger_demo).  Furthermore, after the initial fork the action runs and fails which gives impression that user did something wrong.

Proposal is to make locking optional and since the demo repo uses null resources it won't really be needed, instead we will demonstrate the digger in the most basic from allows you to collaborate on pull requests through several flows such as comment-based flow and see plans or applies directly in the pull request.

In case a user does not supply keys we proceed to perform the terraform action as usual but display a warning to the user saying that locking isn't being actioned since no keys were supplied and that they need to supply cloud account keys in order to allow locking.

vedant-rox:
On the current `develop` branch (the repo's default branch), locking appears to already be optional.

- `docs/ce/howto/disable-locking.mdx` documents disabling it with `pr_locks: false`.
- `docs/ce/reference/digger.yml.mdx` lists `pr_locks` as a top-level config option.
- `libs/digger_config/yaml.go` defines `PrLocks` in config.
- `backend/controllers/github_comment.go` only performs PR locking when `config.PrLocks` is enabled.
- `action.yml` also still carries the deprecated `disable-locking` input for backward compatibility.

Given the current code and docs, this seems safe to close unless the issue is meant to track a narrower onboarding/docs concern beyond optional locking itself.

PR #278, opened 2023-05-19
feat: add support for disabling locking with NoOp lock provider

duoctranth:
This pull request introduces the ability to disable locking functionality by adding a new action input called `disable-locking`. When set to `true`, it utilizes the NoOp lock provider, which performs no actual locking operations.

The logic code has been implemented to support this new feature, ensuring the appropriate environment is utilized based on the `disable-locking` input value.

For more details, please refer to the linked issue.

Linked Issue: #99

### DIFF
--- a/pkg/utils/locking.go
+++ b/pkg/utils/locking.go
@@ -36,6 +36,21 @@ type Lock interface {
 	GetLock(resource string) (*int, error)
 }
 
+type NoOpLock struct {
+}
+
+func (noOpLock *NoOpLock) Lock(transactionId int, resource string) (bool, error) {
+	return true, nil
+}
+
+func (noOpLock *NoOpLock) Unlock(resource string) (bool, error) {
+	return true, nil
+}
+
+func (noOpLock *NoOpLock) GetLock(resource string) (*int, error) {
+	return nil, nil
+}
+
 type ProjectLock interface {
 	Lock(prNumber int) (bool, error)
 	Unlock(prNumber int) (bool, error)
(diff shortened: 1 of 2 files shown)


pkg/utils/locking.go
```diff
@@ -77,7 +92,9 @@ func (projectLock *ProjectLockImpl) Lock(prNumber int) (bool, error) {
 		return false, err
 	}
 
-	if lockAcquired {
+	_, isNoOpLock := projectLock.InternalLock.(*NoOpLock)
```
Spartakovic, line 95: I think this kinda defeats the purpose of having polymorphic lock if we need to do a type check.

duoctranth: Thank you for your feedback. Regarding the `NoOpLock`, would it be possible to add an additional return parameter called `lockIgnored` to the `Lock` interface instead of using this alternative approach?
I'm open to exploring other ideas as well. Please let me know if you have any suggestions or recommendations.

duoctranth:
> I wonder if it wouldn't be more straightforward instead of having no-op lock, just make a decision about locking on DiggerExecutor level. No-op locking and type checking a level higher look more like workarounds
Thank you for your suggestion. I understand your point about making the decision about locking on the `DiggerExecutor` level instead of using a no-op lock. While it's true that using no-op locking and type checking at a higher level can be seen as workarounds, the reason we opted for this approach is to keep the code changes minimal and provide the flexibility to customize the locking behavior in the future if needed.

By using the no-op lock and type checking, we can easily switch between enabling and disabling locking without modifying the DiggerExecutor logic. This allows us to maintain a clear separation between the locking mechanism and the executor logic. Additionally, it provides an opportunity for customization by allowing different messages to be displayed later on.

However, I appreciate your input, and if you think there is a more straightforward approach that aligns with the project's goals and architecture, I'm open to discussing it further. Let's explore all options and come up with the best solution together. Feel free to share any specific ideas or concerns you have regarding the locking implementation.

motatoes (approved):

PR #291, opened 2023-05-23
add unstable to mergeable

motatoes:
Unstable means: "Mergeable with non-passing commit status." So it should be in the map of mergeable states

https://docs.github.com/en/github-ae@latest/graphql/reference/enums#mergestatestatus

### DIFF
--- a/pkg/github/github.go
+++ b/pkg/github/github.go
@@ -90,6 +90,7 @@ func isMergeableState(mergeableState string) bool {
 	// https://docs.github.com/en/github-ae@latest/graphql/reference/enums#mergestatestatus
 	mergeableStates := map[string]int{
 		"clean":     0,
+		"unstable":  0,
 		"has_hooks": 1,
 	}
 	_, exists := mergeableStates[strings.ToLower(mergeableState)]


fleroux514:
I'm trying to lookup a use-case that would result with this status but its a bit obscure. Even chat-GPT doesn't seem to know about it :).

Did you have a specific scenario in mind regarding this fix? In doubt, I personally would prefer having a more strict behavior for now and add `unstable` when we can verify that there is indeed a bug and its needed, then the other way around that might result in and unwanted PR merge when running  `apply`.

The function `isMergeableState()` prints the status at line 98 when returning `false`. And in the worst case, users could merge the PR manually if needed after running `apply`.

motatoes:
hey @fleroux514 thanks for looking into this I did run into this while setting up digger on a private repo, although all the checks were successful the digger apply command failed with the error "PR is not in a mergable state" while the merge button was active and I could infact merge it. However the apply was no longer possible. I am going to attempt reproducing this in a public repo

motatoes:
@fleroux514 here is a reproduction of the issue

pull request: https://github.com/motatoes/digger-test-mergeable/pull/1

action in unstable state: https://github.com/motatoes/digger-test-mergeable/actions/runs/5059997314

you will see that the pull request has 1 commit which fails plan due to invalid terraform, now the next commit has succeeded te terraform plan phase but I am now unable to apply due to it being in an "unstable" state

fleroux514:
Thanks @motatoes. I find it to be a weird behavior from Github but based on this test I think we can proceed with this change.

However it is not the reason why `digger apply` is not working. It should only affect if the PR gets merged or not (the apply command is supposed to have worked)

fleroux514 (approved):

frank-bee:
> @fleroux514 here is a reproduction of the issue
> 
> pull request: [motatoes/digger-test-mergeable#1](https://github.com/motatoes/digger-test-mergeable/pull/1)
> 
> action in unstable state: https://github.com/motatoes/digger-test-mergeable/actions/runs/5059997314
> 
> you will see that the pull request has 1 commit which fails plan due to invalid terraform, now the next commit has succeeded te terraform plan phase but I am now unable to apply due to it being in an "unstable" state (although its perfectly mergeable)

FYI : I also have this issue ( we are working it github.com enterprise with private repos)

frank-bee:
I tried this again with 
```
 - name: digger run
        uses: diggerhq/digger@main"
```
Still having the issue

frank-bee:
... sorry , I see , that the github action with this fix is not released yet. got it

motatoes:
hey @frank-bee it should be fixed in `@v0.1.21` :)

PR #312, opened 2023-06-02
fix: check that there at least one command run

Spartakovic:


pkg/digger/digger.go
```diff
@@ -141,12 +141,17 @@ func RunCommandsPerProject(commandsPerProject []ProjectCommand, repoOwner string
...
 	for _, success := range appliesPerProject {
 		if !success {
 			allAppliesSuccess = false
 		}
 	}
-	return allAppliesSuccess, nil
+
+	atLeastOneCommandRun := len(commandsPerProject) > 0
+
+	return allAppliesSuccess && atLeastOneCommandRun, nil
```
motatoes: this looks like a hack, you are returning false that the runs were successful which is misleading. Why not do this check explicitly in main.go ? It will be much clearer there

Spartakovic: not sure what do you mean? There is nothing about successful runs here, only about applies

motatoes: In https://github.com/diggerhq/digger/blob/9efa3a7d63c13ffbd036763c4c946756a00b5b1f/cmd/digger/main.go#L86 the result of the function is a boolean indicating if all commands are successful or not. You are returning false if there was no projects showing that the applies were not successful. My suggestion is to keep the check  len(commandsPerProject) > 0 directly in main.go where we check for automerge if

Spartakovic: there is nothing saying that all commands are successful? all variables are explicitely called like allAppliesSuccess, not allCommandSuccess

Spartakovic: if I return allAppliesSuccess when there were no projects changes, it's gonna be misleading too

Spartakovic: I understand the concern tho and will accept a better solution

Spartakovic: what I can do is to have len(commandsPerProject) > 0 check in main.go and just do an early exit I guess, not just skipping automerge. But then I will need to do it in several places, as well, here at least it's isolated in one

motatoes: I think checking earlier than the function and exiting with success message is better, "no commands to run, exiting". returning false from function indicates that there were there were some commands and they, at least from my perspective it does not make sense

motatoes (approved):

PR #316, opened 2023-06-05
Add more digger config validators

veziak:


fleroux514:
Just a thought, but I though this library might be very useful to validate the digger.yml config and provide useful error feedback to users:  https://github.com/go-playground/validator

https://medium.com/tunaiku-tech/go-validator-v10-c7a4f1be37df

PR #331, opened 2023-06-07
split DiggerConfig into two parts, one for yaml parsing and one for config itself

veziak:


pkg/configuration/digger_config.go
```diff
@@ -243,6 +279,94 @@ func defaultWorkflow() *Workflow {
...
+	}
+	return result
+}
+
+func copyEnvVars(envVars *EnvVarsYaml) *EnvVarsConfig {
+	result := EnvVarsConfig{}
+	result.State = make([]EnvVarConfigConfig, len(envVars.State))
+	result.Commands = make([]EnvVarConfigConfig, len(envVars.Commands))
+
+	for i, s := range envVars.State {
```
Spartakovic: envVars can be null I think?

pkg/configuration/digger_config.go
```diff
@@ -243,6 +279,94 @@ func defaultWorkflow() *Workflow {
...
+		item := EnvVarConfigConfig{
+			s.Name,
+			s.ValueFrom,
+			s.Value,
+		}
+		result.State[i] = item
+	}
+	for i, s := range envVars.Commands {
```
Spartakovic: same here

pkg/configuration/digger_config.go
```diff
@@ -243,6 +279,94 @@ func defaultWorkflow() *Workflow {
...
+	}
+
+	return &result
+}
+
+func copyStage(stage *StageYaml) *models.Stage {
+	result := models.Stage{}
+	result.Steps = make([]models.Step, len(stage.Steps))
+
+	for i, s := range stage.Steps {
```
Spartakovic: stage can be null too

veziak:
> all "Config" structs should move to core/models package I think.

should we keep yaml and config structures together in configuration package ?

pkg/configuration/converters.go
```diff
@@ -0,0 +1,169 @@
...
+		result.Steps[i] = item
+	}
+	return &result
+}
+
+func copyWorkflowConfiguration(config *WorkflowConfigurationYaml) *WorkflowConfiguration {
+	result := WorkflowConfiguration{}
+	result.OnPullRequestClosed = make([]string, len(config.OnPullRequestClosed))
+	result.OnPullRequestPushed = make([]string, len(config.OnPullRequestPushed))
+	result.OnCommitToDefault = make([]string, len(config.OnCommitToDefault))
```
Spartakovic: do you need to do this given you just override them below?

pkg/configuration/converters.go
```diff
@@ -0,0 +1,169 @@
...
+		plan := copyStage(w.Plan)
+		apply := copyStage(w.Apply)
+		configuration := copyWorkflowConfiguration(w.Configuration)
+		item := Workflow{
+			envVars,
+			plan,
+			apply,
+			configuration,
+		}
+		copier.Copy(&w, &item)
```
Spartakovic: do you need to do both Copy and manuall assigning variables?

veziak: oops, I think it's merge mistake, that copier lib didn't even work for me

pkg/configuration/converters.go
veziak: yep, seems it's not needed

PR #369, opened 2023-06-15
fix: Support case insensitive digger commands

### LINKED ISSUE #344: Support case insensitive digger commands
Requested by user:

Digger commentOps currently only work with exact lowercase "digger plan", "digger apply" etc.

Digger should be able to also support case insensitve commands such as "Digger plan" and so on since it is more forgiving.

Proposal: make digger case insensitive by default, allow users to make it case sensitive through some parameter passed into composite action `case-sensitve` (defaults to false)

tarunsamanta2k20:
closes #344 

**Issue Description:**

Requested by user:

Digger commentOps currently only work with exact lowercase "digger plan", "digger apply" etc.

Digger should be able to also support case insensitve commands such as "Digger plan" and so on since it is more forgiving.

Proposal: make digger case insensitive by default, allow users to make it case sensitive through some parameter passed into composite action case-sensitve (defaults to false)

### DIFF
--- a/pkg/azure/azure.go
+++ b/pkg/azure/azure.go
@@ -450,11 +450,12 @@ func ConvertAzureEventToCommands(parseAzureContext Azure, impactedProjects []con
 		}
 
 		supportedCommands := []string{"digger plan", "digger apply", "digger unlock", "digger lock"}
+		inputCommands := strings.ToLower(diggerCommand)
 		for _, command := range supportedCommands {
-			if strings.Contains(diggerCommand, command) {
+			if strings.Contains(inputCommands, command) {
 				for _, project := range runForProjects {
 					workspace := project.Workspace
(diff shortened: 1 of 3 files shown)


motatoes (changes requested):
Please also make the change in gitlab flow: https://github.com/diggerhq/digger/blob/ac7c2ef6a5f0dd8dc5d0a601db48481ac0aa95d2/pkg/gitlab/gitlab.go#L325

tarunsamanta2k20:
@veziak next time you should see the assignees  of an issue before opening a PR. i was assigned and worked on it. why did you open a PR then?

veziak:
Hi @tarunsamanta2k20 , apologies! I should have communicated it better. We wanted that feature to be included into the release we created yesterday and it was easier to create a new PR.

PR #372, opened 2023-06-15
Support org level rbac

motatoes:


pkg/policy/policy.go
```diff
@@ -22,43 +23,91 @@ type DiggerHttpPolicyProvider struct {
...
 }
 
-func (p NoOpPolicyChecker) Check(_ string, _ string, _ interface{}) (bool, error) {
+func (p NoOpPolicyChecker) Check(_ string, _ string, _ string, _ interface{}) (bool, error) {
 	return true, nil
 }
 
-func (p *DiggerHttpPolicyProvider) GetPolicy(namespace string, projectName string) (string, error) {
-	// fetch RBAC policies from Digger API
+func (p *DiggerHttpPolicyProvider) getPolicyForOrganisation(organisation string, projectName string) (string, *http.Response, error) {
```
Spartakovic: don't need a projectname probably

veziak: fixed

pkg/github/github.go
```diff
@@ -30,10 +30,29 @@ type GithubService struct {
 	Owner    string
 }
 
+func (svc *GithubService) GetUserTeams(organisation string, user string) ([]string, error) {
+	teamsResponse, _, _ := svc.Client.Teams.ListTeams(context.Background(), organisation, nil)
```
Spartakovic: should probably handle errors here

pkg/github/github.go
```diff
@@ -30,10 +30,29 @@ type GithubService struct {
...
+		teamMembers, _, _ := svc.Client.Teams.ListTeamMembersBySlug(context.Background(), organisation, *team.Slug, nil)
+		for _, member := range teamMembers {
+			if *member.Login == user {
+				teams = append(teams, *team.Name)
+				break
+			}
+		}
+	}
+
+	for _, v := range teams {
```
Spartakovic: remove printing

pkg/github/github.go
veziak: done

pkg/github/github.go
veziak: done

pkg/digger/digger.go
```diff
@@ -90,6 +104,7 @@ func RunCommandsPerProject(
 				if err != nil {
 					log.Printf("Error publishing comment: %v", err)
 				}
+				log.Println(msg)
```
Spartakovic, line 107: remove

veziak: that message is actually useful :) it duplicates what we post as comment, but it make clear why job failed in logs

Spartakovic: ah alright, we probably need to sort out usages of log, fmt, and println at some point

PR #404, opened 2023-06-23
Priority order of policies

Spartakovic:
* Fix priority order of policies. First we need to check if there is a particular policy for a project and only then check general org policy

* If no policy exist, we allow anything


veziak (approved):
lgtm

PR #415, opened 2023-06-28
Add reporting strategies

Spartakovic:
image

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -89,7 +89,7 @@ inputs:
   reporting-strategy:
     description: 'comments_per_run or latest_run_comment, anything else will default to original behavior of multiple comments'
     required: false
-    default: 'multiple-comments'
+    default: 'comments_per_run'
 outputs:
   output:
     value: ${{ steps.digger.outputs.output }}


action.yml
```diff
@@ -86,6 +86,10 @@ inputs:
     description: Post plans as one comment
     required: false
     default: 'false'
+  reporting-strategy:
+    description: 'comments_per_run or latest_run_comment, anything else will default to original behavior of multiple comments'
```
motatoes, line 90: @Spartakovic I think that the default behaviour should be `comments_per_run` since multiple_comments is too noisy for all people

Spartakovic: don't mind that, only left it so people aren't surprised by the new behaviour

motatoes: lets change it, there is already a manual version update so there shouldn't be surprises I think, we can advise that in the release notes too
in many ways this is much better than the previous behaviour

motatoes (changes requested):

motatoes (approved):

PR #420, opened 2023-06-29
support configurable digger organisation for digger rbac

motatoes:
- support configurable digger organisation for digger rbac
- improvements to allow for setting of organisation
-

### DIFF
--- a/pkg/core/policy/policy.go
+++ b/pkg/core/policy/policy.go
@@ -1,9 +1,11 @@
 package policy
 
+import "digger/pkg/ci"
+
 type Provider interface {
 	GetPolicy(namespace string, projectname string) (string, error)
 }
 
 type Checker interface {
-	Check(SCMOrganisation string, namespace string, projectname string, command string, requestedBy string) (bool, error)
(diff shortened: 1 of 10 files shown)


pkg/core/policy/policy.go
```diff
@@ -5,5 +5,5 @@ type Provider interface {
 }
 
 type Checker interface {
-	Check(organisation string, namespace string, projectname string, command string, requestedBy string) (bool, error)
+	Check(githubOrganisation string, namespace string, projectname string, command string, requestedBy string) (bool, error)
```
Spartakovic: shouldn't rename it

pkg/digger/digger.go
```diff
@@ -75,15 +75,15 @@ func RunCommandsPerProject(
 ) (bool, bool, error) {
 	appliesPerProject := make(map[string]bool)
 
-	organisation := strings.Split(projectNamespace, "/")[0]
```
Spartakovic, line 78: this is generic function called by other CI providers, so shouldn't rename it either

Issue #399, opened 2023-06-21
Manual merge of a PR doesn't run `digger apply` properly


fleroux514:
Using digger@v0.1.25

My digger workflow is setup with 
```
    workflow_configuration:
      on_pull_request_pushed: [digger plan]
      on_pull_request_closed: [digger unlock]
      on_commit_to_default: [digger apply]
```

I manually merged the PR and logs show the following:

```
2023/06/21 22:03:50 Following projects are impacted by pull request #571
artifact_registries_prod
gke_shared_cluster_dev
gke_shared_cluster_non-prod
gke_shared_cluster_prod
nlu_speech_science_dev
nu_workforce_dev
nu_workforce_non_prod
nu_workforce_prod
personal_projects_dev
website_corpo_prod
GitHub event processed successfully
GitHub event converted to commands successfully
2023/06/21 22:03:50 Following commands are going to be executed:
project: artifact_registries_prod: commands: "digger apply", 
project: gke_shared_cluster_dev: commands: "digger apply", 
project: gke_shared_cluster_non-prod: commands: "digger apply", 
project: gke_shared_cluster_prod: commands: "digger apply", 
project: nlu_speech_science_dev: commands: "digger apply", 
project: nu_workforce_dev: commands: "digger apply", 
project: nu_workforce_non_prod: commands: "digger apply", 
project: nu_workforce_prod: commands: "digger apply", 
project: personal_projects_dev: commands: "digger apply", 
project: website_corpo_prod: commands: "digger apply", 
Commands executed successfully
Digger finished successfully
```

But the apply did not actually happen  (even the logs don't show the apply results)

ZIJ:
Thanks for reporting @fleroux514 ! we'll look into it

ZIJ:
Potential alternative solution: add an `auto-apply` flag to apply after merge. BUT that'd interfere with the workflow configuration (they cannot work together). OR we could check if there's custom workflow, we ignore the mergability check.

veziak:
Hi @fleroux514, we just re-release a new version that should fix the issue

fleroux514:
As far as I'm concerned we can close this.

PR #422, opened 2023-06-30
fix: verify if PR was manually merged with auto-merge config

fleroux514:
Relates to #399 

Complement to fix made in #401 

Fixing the use case when manually merging a PR with the following digger.yml config:
- auto-merge: true
- `on_commit_to_default: [digger apply]`

Digger would run the `apply` cmd and then verify if the PR is mergeable (for the auto-merge config), which would result in a failure.

### DIFF
--- a/pkg/digger/digger.go
+++ b/pkg/digger/digger.go
@@ -297,5 +297,7 @@ func MergePullRequest(ciService ci.CIService, prNumber int) {
 		if err != nil {
 			log.Fatalf("failed to merge PR, %v", err)
 		}
+	} else {
+	   log.Printf("PR is already merged, skipping merge step")
 	}
 }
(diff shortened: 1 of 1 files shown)


pkg/digger/digger.go
```diff
@@ -264,28 +265,37 @@ func SortedCommandsByDependency(project []models.ProjectCommand, dependencyGraph
...
+
+		if !prIsMergeable {
+			log.Fatalf("PR is not mergeable")
+		}
+
+		err = ciService.MergePullRequest(prNumber)
+		if err != nil {
+			log.Fatalf("failed to merge PR, %v", err)
+		}
 	}
```
motatoes, line 302: maybe worth adding a 
```suggestion
	} else {
	   log.Printf("PR is already merged, skipping merge step")
	}
```

motatoes (changes requested):

motatoes (approved):

PR #423, opened 2023-06-30
fix: fail workflow when a project lock is acquired by another PR

fleroux514:
In a mono-repo with 10 projects, when a lock cannot be acquired for 1 particular project, the workflow would still post the other 9 plan results in a comment and end successfully as if all was good, which is very misleading.

Therefore making a change to fail the workflow when a project lock cannot be acquired.


fleroux514:
Just restested in latest develop and the new reports show explicitly when a lock was not acquired. Thus closing this change since its not valid anymore.

image

Issue #443, opened 2023-07-12
Auto-detect projects configuration


Spartakovic:
For very simple project structures, it would nice to have digger.yml-less flow. Digger could traverse repo and guess project structure and auto-generate configuration

motatoes:
I think the basic implementation here was implemented in #444 I'm going to close this for the time being

PR #444, opened 2023-07-12
add auto-detection of projects in case of missing digger.yml

Spartakovic:
https://github.com/diggerhq/digger/issues/443


veziak:
I would move auto detection tests to a separate file, because it's important feature that probably will be improved and grow a bit more

veziak (approved):
lgtm

motatoes (approved):

PR #446, opened 2023-07-14
AWS envprovider changes

HamzaMasood1:
I hope this is a good cleanup of this file. In this way, it is easier in terms of maintainability and extendability of the env vars.

### DIFF
--- a/pkg/aws/envprovider/envprovider.go
+++ b/pkg/aws/envprovider/envprovider.go
@@ -38,14 +38,14 @@ func (e *EnvProvider) Retrieve() (credentials.Value, error) {
 	e.retrieved = false
 
 	//assign id from env vars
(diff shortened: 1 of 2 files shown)


veziak (commented):
@HamzaMasood1 could you please add a new file with unit tests for envprovider.go. It will make it easier to understand the behaviour

motatoes:
looks like parallel tests are not possible to set env variables @HamzaMasood1, we need to fix [those tests](https://github.com/diggerhq/digger/actions/runs/5557695986/jobs/10186762698?pr=446):

```
panic: testing: t.Setenv called after t.Parallel; cannot set environment variables in parallel tests [recovered]
	panic: testing: t.Setenv called after t.Parallel; cannot set environment variables in parallel tests
```

veziak:
looks like tests are failing because of this:
2023-07-17T13:10:34.2180999Z panic: testing: t.Setenv called after t.Parallel; cannot set environment variables in parallel tests [recovered]

PR #448, opened 2023-07-17
Env promotions

veziak:


pkg/core/terraform/tf.go
```diff
@@ -155,33 +157,70 @@ func (sw *StdWriter) GetString() string {
...
+func (tf Terraform) formatTerraformWorkspaces(list string) string {
 
 	list = strings.TrimSpace(list)
 	char_replace := strings.NewReplacer("*", "", "\n", ",", " ", "")
 	list = char_replace.Replace(list)
 	return list
 }
 
+func copyGitHubVars() (map[string]string, error) {
+	githubVars := os.Getenv("GITHUB_VARS")
```
Spartakovic: there shouldn't be anything provider specific inside of "core"

pkg/core/terraform/tf.go
```diff
@@ -155,33 +157,70 @@ func (sw *StdWriter) GetString() string {
...
+				return nil, err
+			}
+		}
+	}
+	return result, nil
+}
+
 func (tf Terraform) Plan(params []string, envs map[string]string) (bool, string, string, error) {
-	workspaces, _, _, err := tf.runTerraformCommand("workspace", envs, "list")
+	githubVars, _ := copyGitHubVars()
```
Spartakovic: same as before, also why do we even need this?

pkg/core/terraform/tf.go
```diff
@@ -155,33 +157,70 @@ func (sw *StdWriter) GetString() string {
...
+	}
+
+	workspaces, _, _, err := tf.runTerraformCommand("workspace", nil, "list")
 	if err != nil {
 		return false, "", "", err
 	}
 	workspaces = tf.formatTerraformWorkspaces(workspaces)
 	if strings.Contains(workspaces, tf.Workspace) {
-		_, _, _, err := tf.runTerraformCommand("workspace", envs, "select", tf.Workspace)
+		_, _, _, err := tf.runTerraformCommand("workspace", nil, "select", tf.Workspace)
```
Spartakovic: I would check that you need to do this again

pkg/core/terraform/tf.go
veziak: similar to secrets we need to get json with all github vars and set env vars accordingly, another option would be to map every var individually in github workflow's config

pkg/core/terraform/tf.go
veziak: yeah probably I need to move it a few levels up

pkg/core/terraform/tf.go
Spartakovic: why do we need all vars?

Spartakovic: for promotion they will just need to set one variable right? ENV_NAME or something. Setting all variables is an overkill

pkg/core/terraform/tf.go
veziak: it seems it can be reverted back, it didn't work when I was mapping envs to vars, but I removed that code so it should be ok

pkg/core/terraform/tf.go
veziak: that would allow to map github vars to terraform vars individually without tfvars file

Spartakovic: hmm, should we at least limit them, to something like "if starts with TF_"

Spartakovic (approved):

PR #449, opened 2023-07-17
Introduce manual execution mode

Spartakovic:


pkg/ci/ci.go
```diff
@@ -1,6 +1,6 @@
 package ci
 
-type CIService interface {
+type PullRequestService interface {
```
veziak, line 3: should we rename this interface ? looks like we need to extend it ?

Spartakovic: we don't need to extend it, I split it in two. PullRequest and Org services. Because manual mode doesn't care about anything related to pull requests, and only needs org information to do policy check

Spartakovic: Github service (and others) actually implement both

Spartakovic: I didn't _have_ to do it to make everything work, but it feels right

veziak: nice!

motatoes (approved):

PR #454, opened 2023-07-18
Fix for opa regressions

motatoes:
we were not properly composing urls for the case when user sets url to "https://cloud.uselemon.cloud/" we formed invalid urls with double /

now we use golang url library to compose the urls

also we improve the readability of `getPolicy` function by adding comments and splitting out the if statements to be more clear and readable

validated to be working in new pull request of test repository: https://github.com/digger-tests/digger-demo-opa/pull/8


veziak (approved):
lgtm

PR #469, opened 2023-07-26
feat: create text file in project root repository when plan returned changes

fleroux514:
This allows to add logic in the digger workflow based on if the plan returned changes or not.

For instance, to run `checkov` only when the plan returned changes:

```
workflows:
  default:
    plan:
      steps:
        - init
        - plan
        - run: if [ -f "isNonEmptyPlan.txt" ]; then checkov --directory .; fi
```

The `-no-color` plan default argument was also added in this PR to fix a bug in `cleanupTerraformOutput()` function when `nonEmptyOutput` is false. The following logic would not work because of color characters around the `No changes` substring:

```
	} else if nonEmptyOutput {
		start = "Terraform will perform the following actions:"
	} else {
		start = "No changes. Your infrastructure matches the configuration."
	}

```

### DIFF
--- a/pkg/core/execution/execution.go
+++ b/pkg/core/execution/execution.go
@@ -138,6 +138,15 @@ func (d DiggerExecutor) Plan() (bool, bool, string, string, error) {
 			planArgs = append(planArgs, step.ExtraArgs...)
 			nonEmptyPlan, stdout, stderr, err := d.TerraformExecutor.Plan(planArgs, d.CommandEnvVars)
 			isNonEmptyPlan = nonEmptyPlan
+			if isNonEmptyPlan {
+				nonEmptyPlanFilepath := strings.Replace(d.PlanPathProvider.LocalPlanFilePath(), d.PlanPathProvider.PlanFileName(), "isNonEmptyPlan.txt", 1)
+				file, err := os.Create(nonEmptyPlanFilepath)
+				if err != nil {
+					return false, false, "", "", fmt.Errorf("unable to create file: %v", err)
+				}
+				defer file.Close()
+			}
+
 			if err != nil {
 				return false, false, "", "", fmt.Errorf("error executing plan: %v", err)
 			}


Spartakovic:
why can't we check the plan file itself?

fleroux514:
Do you mean the content of the plan file? It's binary, and the content by itself represents some infrastructure configuration, but not the diff vs the current infrastructure state.

Also in the digger cmds, seems like we cannot use pipes with multiple commands.

How would you do it?

Spartakovic:
that's true my bad, for some reason I thought it's gonna be actualy empty file.

We have now enabled -detail-exitcode by default. Will it help with your original suggestion?

fleroux514:
Realised after looking at the code that `isNonEmptyPlan` variable is what was really needed, which is why I based my proposal on it.

From the workflow itself, I cannot read that the `plan` command returned 2, and I don't think there is an alternative to either my proposal or setting an env variable. Using an env variable would be a bit more complicated since it would need to be reset between each project in a mono-repo.

Maybe digger could add a `run_when_plan_changes` step that would rely on this file to execute the cmd instead of using bash like my example
```
workflows:
  default:
    plan:
      steps:
        - init
        - plan
        - run_when_plan_changes:  checkov --directory .
```

pkg/core/execution/execution.go
```diff
@@ -134,10 +134,19 @@ func (d DiggerExecutor) Plan() (bool, bool, string, string, error) {
 			}
 		}
 		if step.Action == "plan" {
-			planArgs := []string{"-out", d.PlanPathProvider.PlanFileName(), "-lock-timeout=3m"}
+			planArgs := []string{"-no-color", "-out", d.PlanPathProvider.PlanFileName(), "-lock-timeout=3m"}
```
Spartakovic: we already add -no-color by default now

fleroux514: reverted

Spartakovic (approved):

PR #473, opened 2023-07-31
feat: extend drift detection notification url types

fleroux514:
Tested the new drift detection feature with a Teams webhook (instead of Slack) and it worked perfectly fine (great feature!)

The only different is how to specify the bangbang emoji prefix included in the notification message so I suggest to remove the `slack` reference to all variables and adding a `drift-detection-notification-url-type` input variable tu support to multiple channel types.

I wasn't able to easily include the "bangbang" emoji using text for teams so I didn't add the "teams" as part of `drift-detection-notification-url-type` supported values.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -105,10 +105,14 @@ inputs:
     description: 'project name for digger to run in case of manual mode'
     required: false
     default: ''
-  drift-detection-slack-notification-url:
-    description: 'drift-detection slack notification url'
+  drift-detection-notification-url:
+    description: 'drift-detection notification url'
     required: false
     default: ''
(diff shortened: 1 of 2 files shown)


fleroux514:
Understood. Just added this quick fix in case you were about to release this feature.

motatoes:
@fleroux514 really grateful that you managed to test it so early and make it work for teams! 😄

motatoes:
Closing this for now while we make a decision on the notification channels! Thanks again for this idea and investing time to create a PR for it, we will let you know what we end up doing

PR #480, opened 2023-08-04
add reporting of projects and runs

Spartakovic:


pkg/policy/policy.go
```diff
@@ -253,9 +253,8 @@ type DiggerPolicyChecker struct {
 }
 
 func (p DiggerPolicyChecker) CheckAccessPolicy(ciService ci.OrgService, SCMOrganisation string, SCMrepository string, projectName string, command string, requestedBy string) (bool, error) {
-	// TODO: Get rid of organisation if its not needed
-	organisation := p.PolicyProvider.GetOrganisation()
-	policy, err := p.PolicyProvider.GetAccessPolicy(organisation, SCMrepository, projectName)
+
+	policy, err := p.PolicyProvider.GetAccessPolicy(SCMOrganisation, SCMrepository, projectName)
```
motatoes, line 257: this has to be `organisation` in fact since the digger backend only cares about digger organisation and should be unaware of the SCM specific organisation, that's the point

motatoes (approved):

PR #481, opened 2023-08-07
fix default project name in AutoDetectDiggerConfig

veziak:


motatoes:
please write some tests for this edge case

PR #482, opened 2023-08-08
Add debug config WIP

veziak:


pkg/core/execution/execution.go
```diff
@@ -134,7 +134,7 @@ func (d DiggerExecutor) Plan() (bool, bool, string, string, error) {
 			}
 		}
 		if step.Action == "plan" {
-			planArgs := []string{"-out", d.PlanPathProvider.PlanFileName(), "-lock-timeout=3m"}
+			planArgs := []string{"-out", d.PlanPathProvider.LocalPlanFilePath(), "-lock-timeout=3m"}
```
veziak, line 137: not if this change is going to break something else

PR #554, opened 2023-09-06
fix apply failures when checkov is configured

fleroux514:
Error seems to have been introduced in #449 which was included in [v0.1.30](https://github.com/diggerhq/digger/releases/tag/v0.1.30)

digger apply results with this error when checkov is configured

image


veziak:
I think it's ok for now, but we will need to make it CI agnostic later, that code shouldn't depend on github

veziak (approved):
lgtm

PR #555, opened 2023-09-07
Fix a bug that do not allow to run digger if aws credentials are not provided

veziak:


pkg/locking/locking_test.go
```diff
@@ -32,6 +32,16 @@ func TestLockingTwiceThrowsError(t *testing.T) {
...
 	assert.NoError(t, err2)
 }
+
+func TestGetLock(t *testing.T) {
+	lock, err := GetLock()
+	if err != nil {
+		print(err)
+
+	}
+	print(lock)
```
motatoes, line 45: NIT: should we remove these prints or keep them ?

pkg/locking/locking.go
```diff
@@ -198,26 +198,33 @@ func GetLock() (locking.Lock, error) {
...
-			Config: awssdk.Config{
-				Region:      awssdk.String(awsRegion),
-				Credentials: credentials.NewCredentials(&envprovider.EnvProvider{}),
-			},
-		})
+
+		// https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html
+		options := session.Options{SharedConfigState: session.SharedConfigEnable}
+		if awsProfile != "" {
+			options = session.Options{
```
motatoes, line 222: maybe another print here saying "using keyless credentials" would help in debugging ?

pkg/locking/locking_test.go
veziak: added todo to implement this test properly later

motatoes (approved):

PR #584, opened 2023-09-19
chore: Use ldflags to automatically set the version to the Git tag

kevinmichaelchen:
https://docs.github.com/en/webhooks/webhook-events-and-payloads#release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,10 @@ jobs:
         uses: actions/checkout@v4
       - name: Build
         run: |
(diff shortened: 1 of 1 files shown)


veziak:
Hi @kevinmichaelchen thanks for the PR!, the path to the variable looks a bit wrong but I created another PR that should fix it https://github.com/diggerhq/digger/pull/583

kevinmichaelchen:
Thanks @veziak — I'll pull the latest release and test things out in my **[Tea package](https://github.com/teaxyz/pantry/pull/3358)**

veziak:
@kevinmichaelchen we are going to create a new release next week, until then you probably will not be able to try the fix. Please let me know if it's urgent for you, we can speed it up if needed.

Issue #667, opened 2023-10-23
Update telemetry in line with token auth / Github App


ZIJ:
We currently state in the readme that "No sensitive or personal / identifyable data is logged. You can see what is tracked in pkg/usage/usage.go"

This made sense for a "headless" action but now that we have a dashboard and are implementing Github App and token-based auth, there's just no point, we'll know the github userid anyways. So sending the hash is redundant - it just makes analytics harder without adding any extra anonymity to the user.

PR #668, opened 2023-10-23
Make telemetry explicit

ZIJ:
Addresses #667


motatoes:
looks good to me, also we would need to not hash our users in the lambda to ignore events of those correctly as well https://github.com/diggerhq/digger-usage-lambda/blob/main/app.py#L15

motatoes (approved):

PR #731, opened 2023-11-02
Use source version of libs/orchestrator instead of external module that has been archived

ramin:
updates pkg/github/models/models.go to use the local version of lib/orchestrator

If became evident that though the repo https://github.com/diggerhq/lib-orchestrator had been archived because the library had been merged into `diggerhq/digggr` the code was still referencing the external module. I have updated to use the internal and most up to date version of `libs/orchestrator` vs `github.com/digghq/lib-orchestrator`

also removes `github.com/diggerhq/digger/lib-orchestrator` from `go.mod` and `go.sum`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,6 @@ require (
 
 require (
 	github.com/caarlos0/env/v8 v8.0.0
-	github.com/diggerhq/lib-orchestrator v0.0.11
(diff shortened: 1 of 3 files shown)


motatoes:
Thanks! It seems I missed that in my recent PR which merges the libs

motatoes (approved):

PR #746, opened 2023-11-07
split larger slack notifications into multiple messages to prevent code block formatting from breaking

motatoes:


ZIJ (approved):
LGTM - and liking the refactor so that we can have more notification integrations. Issues, webhooks etc

PR #769, opened 2023-11-12
Fix for https://github.com/diggerhq/digger/issues/768

### LINKED ISSUE #768: AzureCI and GitLabCI don't have Panic Recovery like GithubCI and BitbucketCI
Of the four supported CI tools only Github and Bitbucket have code that recovers from a panic, for example from GitHub CI

```
	defer func() {
		if r := recover(); r != nil {
			log.Println(fmt.Sprintf("stacktrace from panic: \n" + string(debug.Stack())))
			err := usage.SendLogRecord(githubActor, fmt.Sprintf("Panic occurred. %s", r))
			if err != nil {
				log.Printf("Failed to send log record. %s\n", err)
			}
			os.Exit(1)
		}
	}()
```
It would be great if this could be written in a manner that makes it not only support all existing CIs but also as new CIs are added this change could cover those as well.

I am going to take a stab at solving this issue myself and submitting the fix as a PR.

ajaxian79:
Moved and refactored panic recovery into the main func and removed from individual ci handlers for a consolidated flow.

Closes #768

### DIFF
--- a/cmd/digger/main.go
+++ b/cmd/digger/main.go
@@ -830,6 +830,10 @@ func main() {
 		logLeader = os.Getenv("CI_PROJECT_NAME")
 		gitLabCI(lock, policyChecker, backendApi, reportStrategy)
 	case digger.Azure:
+		// This should be refactored in the future because in this way the parsing
+		// is done twice, both here and inside azureCI, a better solution might be
+		// to encapsulate it into a method on the azure package and then grab the
+		// value here and pass it into the azureCI call.
 		azureContext := os.Getenv("AZURE_CONTEXT")
 		parsedAzureContext, _ := azure.GetAzureReposContext(azureContext)
 		logLeader = parsedAzureContext.BaseUrl
--- a/cmd/digger/main.go
+++ b/cmd/digger/main.go
@@ -46,17 +46,6 @@ func gitHubCI(lock core_locking.Lock, policyChecker core_policy.Checker, backend
 		usage.SendUsageRecord("", "log", "non github initialisation")
 	}
 
-	defer func() {
-		if r := recover(); r != nil {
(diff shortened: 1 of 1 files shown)


motatoes:
tests were not triggered for some reason so I will be closing and re-opening to try and trigger them

cmd/digger/main.go
```diff
@@ -853,19 +819,39 @@ func main() {
...
 	switch ci {
 	case digger.GitHub:
+		logLeader = os.Getenv("GITHUB_ACTOR")
 		gitHubCI(lock, policyChecker, backendApi, reportStrategy)
 	case digger.GitLab:
+		logLeader = os.Getenv("CI_PROJECT_NAME")
 		gitLabCI(lock, policyChecker, backendApi, reportStrategy)
 	case digger.Azure:
+		azureContext := os.Getenv("AZURE_CONTEXT")
+		parsedAzureContext, _ := azure.GetAzureReposContext(azureContext)
```
motatoes, line 838: My only concern is that in this case the work for parsing payload will be done twice

It is a minor task but we should probably convert those parsers into singletons in the future

Should be OK to merge for now but can we make a note on top of this line with a comment which says something like "TODO: refactor this function to make it not perform double work after calling the `acureCI` function"

motatoes (changes requested):
Could we just add a comment on the azure section for refactoring in the future? And then its good to go 👍

ajaxian79:
@motatoes I will definitely add a comment when I shoot a follow up video this weekend, thanks!  I may even take a stab at refactoring

motatoes (approved):

motatoes:
❤️  thanks for your contribution @ajaxian79

ajaxian79:
> ❤️  thanks for your contribution @ajaxian79 

It was my genuine pleasure and I look forward to contributing more in the future!

PR #777, opened 2023-11-14
Pass plan policy outputs to access policy before apply

ZIJ:
- Plan is now run before every apply
- Policy violations passed to access policy, which is now run TWICE (before every operation, and right before apply)

TODO later:
-  only run if there's an plan policy
- refactor so that access policy only run once. or a different policy?
- use AccessPolicyContext in CheckAccessPolicy


ZIJ:
What happens if there's no access policy at all? Apply will probably still go ahead. So We might need a "default policy" to prevent that. (Mo 16.11.2023)
- default Acess policy allowing everything, unless there are plan policy violations
- default Plan policy just allows everything

motatoes (approved):
Approved but we need to only perform a terraform plan for the apply route _if_ there is a plan policy configured, otherwise it is a redundant call

PR #788, opened 2023-11-15
fix Github's GetChangedFiles() to also return previous filenames

### LINKED ISSUE #630: moving a tf file from a project to another does not trigger a plan in the source project
Tested using digger@v0.1.33

See https://github.com/diggerhq/francios-tests/pull/47 where a file was moved from `dev` project to `non-prod`.

Resulting digger action only planned in the `non-prod`project

image

fleroux514:
when moving files from one project to another, GetChangedFiles() will now include the previous filepath and new filepath causing plans to run in both projects instead of only the project where the file is getting moved to.

Fixes #630

### DIFF
--- a/libs/orchestrator/github/github.go
+++ b/libs/orchestrator/github/github.go
@@ -3,11 +3,12 @@ package github
 import (
 	"context"
 	"fmt"
+	"log"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


motatoes (approved):

motatoes:
Nice catch!

PR #810, opened 2023-11-23
Enforce auth in CLI

ZIJ:


cmd/digger/main.go
```diff
@@ -795,6 +793,7 @@ func main() {
 			HttpClient: http.DefaultClient,
 		}
 	} else {
+		log.Fatalf("DIGGER_TOKEN not specified. Get one at https://cloud.digger.dev")
```
motatoes: can we use `reportErrorAndExit` instead ?

cmd/digger/main.go
```diff
@@ -795,6 +793,7 @@ func main() {
...
 		policyChecker = policy.NoOpPolicyChecker{}
 		backendApi = backend.NoopApi{}
```
motatoes: remove these two lines too

motatoes (changes requested):

motatoes (approved):

PR #948, opened 2023-12-20
Fix Github environment variables for runs in container

chvima:
There is an issue in Github Actions runner, described here:
https://github.com/actions/runner/issues/2058

In summary, if the action is run in container, the variables github.workspace and similar doesn't correspond to a path inside the container. This can be fixed by using environment variables instead.

Tested this fork on my workflows in a container.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -273,11 +273,11 @@ runs:
         NO_BACKEND: ${{ inputs.no-backend }}
         DEBUG: 'true'
       run: |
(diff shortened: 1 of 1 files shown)


motatoes:
Thanks @chvima for the contribution! I'll do a quick test and merge if no issues

motatoes (approved):

chvima:
thank you! looking forward to the next release

PR #1040, opened 2024-01-12
Fix execution when using http backends

sikha-root:
This PR updates the Digger github action so that it doesn't attempt to switch workspaces when it's unnecessary.
(i.e. terraform workspace is already "default", there's no need to attempt to switch to it.)

Reason being is that the [HTTP backend](https://developer.hashicorp.com/terraform/language/settings/backends/http) in Terraform/OpenTofu does not currently support workspaces:

https://github.com/hashicorp/terraform/issues/26797
https://github.com/opentofu/opentofu/issues/317

Attempting to use any workspace commands gives an error:

```
# terraform workspace list
workspaces not supported
```

which in turn was causing Digger to fail:

```
Running command: terraform [workspace list]
Error: exit status 1
Fatal: Error terraform switch to workspace <nil>
Failed to run digger plan command. error running init: exit status 1
Failed to run commands. error while running command: Failed to run digger plan command. error running init: exit status 1
```

### DIFF
--- a/cli/pkg/core/terraform/opentofu.go
+++ b/cli/pkg/core/terraform/opentofu.go
@@ -24,10 +24,12 @@ func (tf OpenTofu) Init(params []string, envs map[string]string) (string, string
 }
 
 func (tf OpenTofu) Apply(params []string, plan *string, envs map[string]string) (string, string, error) {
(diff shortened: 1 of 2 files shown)


motatoes:
Thanks for the PR @sikha-root !

PR #1085, opened 2024-01-25
[Tiny] - Fix image scr in documentation

ilmax:
While reading through the docs, I found this link to be broken. This should fix it

image


ZIJ:
Thanks so much @ilmax !! We are super grateful for this contribution!

With all the support issues and bug fixes and feature requests we are stretched quite thin so have lots of embarrassing little things like this one all over the place that never get the attention they should be getting; community contributions to fix them are especially helpful 🙏

PR #1098, opened 2024-01-31
Use stored plan in pre-apply plan policy checks

ZIJ:
Screenshot 2024-01-31 at 13 06 17


ZIJ:
Confirmed that it still works after removing the duplicate plan run when stored plan does not exist

https://github.com/diggerhq/test-stored-plan/actions/runs/7728245021/job/21068578418

cli/pkg/digger/digger.go
```diff
@@ -359,38 +360,33 @@ func run(command string, job orchestrator.Job, policyChecker policy.Checker, org
...
 		} else {
 
-			// Running plan before apply to check against access policy
-			// TODO only run plan if policy check present. Always running plan is slow.
+			// checking policies (plan, access)
 
-			_, planPerformed, isNonEmptyPlan, _, planJsonOutput, err := diggerExecutor.Plan()
+			storedPlanJson, err := retrievePlanBeforeApply(planStorage, planPathProvider, diggerExecutor)
+			var planPolicyViolations []string
 			if err != nil {
```
motatoes: I don't think we should check for error and proceed regardless, an error retriving plan should abort immediately, there are many reasons for failing to retrive a plan and we are skipping those

We should instead check explicitly that a Noop PlanStorageProvider and in THAT case to skip the violation checks. For now we can check for os.GetEnv("PLAN_UPLOAD_DESTINATION") == "" then skip

and move the retrival to the else part

ZIJ:
Confirmed working after addressing feedback: https://github.com/diggerhq/test-stored-plan/actions/runs/7729644664/job/21073306830

motatoes (approved):

Issue #1094, opened 2024-01-30
No projects impacted comment appended multiple times


ZIJ:
Reported by user A.T.

chvima:
+1
I have switched to using digger via Github Application (and workflow_dispatch in digger) and I'm also getting lot of spam with "No projects impacted". It seems like Github Application is triggered after every:
- label
- comment
- maybe even request review?

Is this configurable?
Ideally to run only on:
- commits
- comments "digger apply" "digger plan"

My workaround is to "disable" the Application and use workflow with custom configuration on pull_request and on issue_comment with conditions, but that's not ideal.

ZIJ:
Again reported today by a different user (M.U.)

ZIJ:
Addressed in #1136

There should now be much less "no projects impacted" comments.

- Comments that do not start with "digger" ignored
- PR events that have no impacted projects do not trigger any messages
- "Digger starting..." comment is only appended in case there are impacted projects

chvima:
Thanks! Tested a few events and it looks clean now.

PR #1136, opened 2024-02-08
Reduce PR comment noise

ZIJ:
There should now be much less "no projects impacted" comments.
- Ignore comments that do not start with "digger"
- Do not report on PR events that have no impacted projects
Also the "Digger starting..." comment is now also appended only in case there are impacted projects.


motatoes (approved):

ZIJ:
Addresses #1094

PR #1147, opened 2024-02-13
do not handle PR closed event

motatoes:
Current orchestrator behaviour attempts to handle PR closed, this should be ignored


motatoes:
Closing since I think that we do need to run unlock on pr closing. A future improvement could be that we unlock on the orchestrator itself after performing cloud based locks

Issue #1179, opened 2024-02-19
Telemetry data is NOT anonymous


vdmkenny:
From the documentation:

https://github.com/diggerhq/digger/blob/develop/docs/reference/digger.yml.mdx#L102

`collect_usage_data	boolean	true	no	allows collecting anonymised usage and debugging data	This flag is only supported in Digger enterprise`

However, not even an attempt is made on the client-side to anonymise data. In the below example, your `GITHUB_ACTOR` [environment variable](https://docs.github.com/en/actions/learn-github-actions/variables) is sent to the analytics server.

https://github.com/diggerhq/digger/blob/94585242ebec061a5c6ab6009b344a7ee27c1adb/cli/cmd/digger/main.go#L38-L45

https://github.com/diggerhq/digger/blob/94585242ebec061a5c6ab6009b344a7ee27c1adb/cli/pkg/usage/usage.go#L20-L28

github user and repository names are sent, which is certainly non-trivial information.

Compounded with the fact that the opt-out flag was removed in https://github.com/diggerhq/digger/pull/1164 (Starting with release 0.4.1) this makes this software very problematic to use in my opinion. 

In my opinion the argument made in https://github.com/diggerhq/digger/issues/1154 that you should pay for the Enterprise Edition if you want the privilege of privacy holds no water.
This is very much a privacy anti-pattern, and I wonder what the GDPR implications of this are.

Can you point me to anything resembling a [Data Processing Agreement](https://gdpr.eu/what-is-data-processing-agreement/?cn-reloaded=1)?

In any case, re-introducing the flag `collect_usage_data` would be a step in the right direction, but a serious attempt at anonymisation or tokenisation should be made, taking in account the [common misconceptions](https://www.edps.europa.eu/system/files/2021-04/21-04-27_aepd-edps_anonymisation_en_5.pdf).

For users of digger 0.4.1 in github actions, a current workaround to disable usage reporting is to add this step in the action:
```
...
    steps:
      - name: Break digger analytics
        run: |
          sudo echo "127.0.0.1 analytics.digger.dev" | sudo tee -a /etc/hosts
      - name: digger run
        uses: diggerhq/digger@v0.4.1
 ...
```

I should add that I am a small contributor to this repository, and that I genuinely prefer this software over similar solutions such as [Atlantis](https://www.runatlantis.io/). This issue is not an attempt at starting any drama, but to address incorrect information in the documentation, and related privacy concerns.

vdmkenny:
[...]

PR #1164, opened 2024-02-16
change usage domain to analytics and ee only

motatoes:
Related to #1154

- We move our analytics domain to analytics.digger.dev
- We now only support `collect_usage_data` to enterprise edition of digger
- Docs update accordingly

PR #1184, opened 2024-02-20
Revert telemetry changes

motatoes:
addresses #1179

- Reverts https://github.com/diggerhq/digger/pull/1164
- Add pseudo-anonymisation by hashing sensitive userIDs with sha-256
- Add documentation page on how to disable telemetry
- Rename flag from `send_usage_data` to `telemetry` to make it more clear
- Add readme section explaining telemetry changes


libs/digger_config/yaml.go
```diff
@@ -11,6 +11,7 @@ type DiggerConfigYaml struct {
 	Projects                []*ProjectYaml               `yaml:"projects"`
 	AutoMerge               *bool                        `yaml:"auto_merge"`
 	Workflows               map[string]*WorkflowYaml     `yaml:"workflows"`
+	Telemetry               *bool                        `yaml:"telemetry,omitempty"`
```
vdmkenny, line 14: Please keep in mind that people who have not read the changelog or the thread about the issues around telemetry might not see this and update their config flag. Since the feature is opt-out, having the old flag set will re-enable telemetry for these users.

motatoes: it will be mentioned in the next release notes and we will do a psa about it

Issue #1215, opened 2024-02-27
remove single cli build and update action.yml to point to the right build from multiarch


motatoes:
In https://github.com/diggerhq/digger/pull/1214 we have introduced multi builds for cli with different architectures. Now we have a redundant workflow file `.github/workflows/cli_release.yml`

We need to remove that file and update the path from action.yml to point to the right build from multiarch list, probably it will be digger-cli-linux-amd64

PR #1226, opened 2024-02-29
Add SHA256 checksums to release artifacts

sikha-root:
This replaces the action used in release workflows with a Go-specific release action, but the build and release process more or less remains the same as before. We effectively only add an additional checksum generation and publish step with this update.

We also remove `cli_release.yml` because it builds `digger-cli-Linux-X64` which is the same as the `digger-cli-linux-amd64` artifact built by `cli_release_multiarch.yml`. If this isn't desired, we can remove that commit on request.

### DIFF
--- a/.github/workflows/backend_release.yml
+++ b/.github/workflows/backend_release.yml
@@ -3,7 +3,8 @@ name: Backend release
 
 "on":
   release:
-    types: [released]
+    types:
+      - 'released'
 
 jobs:
   binary:
@@ -16,10 +17,13 @@ jobs:
             arch: arm
           - os: darwin
             arch: "386"
+
     runs-on: ubuntu-latest
+
     steps:
       - name: Check out repository
         uses: actions/checkout@v4
+
       - name: Build and publish binary artifact to GitHub
         id: build-and-release-binary
         uses: wangyoucao577/go-release-action@8fa1e8368c8465264d64e0198208e10f71474c87  # v1.50
--- a/.github/workflows/backend_release.yml
+++ b/.github/workflows/backend_release.yml
@@ -1,41 +1,38 @@
+---
 name: Backend release
-on:
+
+"on":
   release:
     types: [released]
+
 jobs:
   binary:
     strategy:
       matrix:
-        arch: [arm, arm64, amd64, 386]
+        arch: [arm, arm64, amd64, "386"]
         os: [linux, darwin, freebsd, windows]
         exclude:
           - os: darwin
             arch: arm
           - os: darwin
-            arch: 386
-
+            arch: "386"
     runs-on: ubuntu-latest
     steps:
-      - name: Download Go
-        uses: actions/setup-go@v5
-        with:
-          go-version: 1.21.1
-        id: go
-
       - name: Check out repository
         uses: actions/checkout@v4
-      - name: Build
-        run: |
-          echo "Tag that is going to be used as digger version: ${{ github.event.release.tag_name }}"
-          env GOOS=${{matrix.os}} GOARCH=${{matrix.arch}} CGO_ENABLED=0 go build -ldflags="-X digger/pkg/utils.version=${{ github.event.release.tag_name }}" -o digger-api ./backend
-      - name: Publish linux-x64 exec to github
-        id: upload-release-asset-linux-x64
-        uses: actions/upload-release-asset@v1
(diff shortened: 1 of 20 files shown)


sikha-root:
Example release: https://github.com/317project/digger/releases/tag/v0.4.6

motatoes:
Hey @sikha-root ! Thanks for this contribution. It looks good to me. One thing - we can't yet remove cli_release.yml although it duplicates because the action.yml downloads that specific one explicitly (see https://github.com/diggerhq/digger/issues/1215). We would need to replace it with some os.arch but when running on ubuntu it doesn't actually translate to the same values that go maps to so its a peice of work to figure out in the future. 

TL/DR; not yet ready to remove the cli_release.yml but rest of the changes look good

sikha-root:
@motatoes OK! I need to take care of some things with my git setup but will remove that change once I'm ready.

motatoes:
> @motatoes OK! I need to take care of some things with my git setup but will remove that change once I'm ready.

Cool! Tag me again when ready @sikha-root

motatoes:
@sikha-root Closing for now but would love for you to re-open once the file is removed to get it merged, since I love the functionality of having a sha checksum for artefacts and it would be great to have your name on this PR :)

sikha-root:
@motatoes thanks, I rebased and left out the commit that removed cli_release.yml.

.github/workflows/cli_release_multiarch.yml
```diff
@@ -1,44 +1,44 @@
...
-        run: |
-          echo "Tag that is going to be used as digger version: ${{ github.event.release.tag_name }}"
-          env GOOS=${{matrix.os}} GOARCH=${{matrix.arch}} CGO_ENABLED=0 go build -ldflags="-X digger/pkg/utils.version=${{ github.event.release.tag_name }}" -o digger ./cli/cmd/digger
 
-      - name: Publish linux-x64 exec to github
-        id: upload-release-asset-linux-x64
-        uses: actions/upload-release-asset@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and publish binary artifact to GitHub
```
motatoes, line 29: I'm getting anxoius about this particular change. I realise that the official action is now unmaintained but I also get anxious about using other third party actions. I'm not sure what the best course of action here is but would rather keep this change out of this PR for a future discussion on the best action to use. 

What do you think ?

.github/workflows/backend_release.yml
```diff
@@ -1,41 +1,42 @@
...
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          upload_url: ${{ github.event.release.upload_url }}
-          asset_path: 'digger-api'
-          asset_name: digger-api-${{matrix.os}}-${{matrix.arch}}
-          asset_content_type: application/octet-stream
 
+      - name: Build and publish binary artifact to GitHub
+        id: build-and-release-binary
+        uses: wangyoucao577/go-release-action@8fa1e8368c8465264d64e0198208e10f71474c87  # v1.50
```
motatoes, line 29: same as my other comment below

.github/workflows/cli_release_multiarch.yml
sikha-root: I don't think there's much reason for concern when using third-party actions as long as we're pinning to the specific commit like we're doing so here. The code can't change from underneath you (as opposed to using tags that *could* get redefined). Unless licensing is your concern?

As for keeping this change out of the PR...that's kind of difficult since it technically is the bulk of this PR and iirc I couldn't figure out how to get and upload checksums in separate steps otherwise. But I'm more familiar with Actions now than I was when I opened this PR (using Digger actually is the first time I've used Actions) so maybe I could figure something out. Not sure.

Then again my original incentive for this PR was that I wanted to add download verification to my Ansible role for deploying self-hosted Digger before publishing it but I've recently started using the helm release so it might take a bit before I try to do anything else with this PR.

motatoes: OK, its mainly a security concern. The fact that its a hardcoded commit helps. I also would need to make not to not merge rennovate bumps unless the new version also reviewed. Lets merge it for now until we find a new solution

motatoes (approved):

Issue #1225, opened 2024-02-28
Sub-directories not found in generate_projects.block glob search


ri-roee:
I have the following `digger.yml` config, with the aims of generating a project for each `main.tf` in my root directory. However at runtime digger is only able to find the project at the root of the directory. Is this a bug with the glob search or am I misconfiguring the `include`? Thanks in advance!

```
generate_projects:
  blocks:
    - include: "deployments/aws/my_cluster/**"
      aws_role_to_assume:
        state: <arn1>
        command: <arn2>
```

```
deployments
-- aws
---- my_cluster
------ main.tf
------ sub-dir
-------- main.tf
------ sub-dir2
-------- folder
---------- main.tf
```

motatoes:
It was sort of a feature back then to stop at the first terraform found during traversal. So based on our current logic I expected it to only find your top level terraform and stop traversing further. We did it that way to avoid picking up a nested modules folder as a terraform project. 

Thinking about it again with your current structure in mind, i now think we should add this option to make it a configurable option

ri-roee:
> It was sort of a feature back then to stop at the first terraform found during traversal. So based on our current logic I expected it to only find your top level terraform and stop traversing further. We did it that way to avoid picking up a nested modules folder as a terraform project.
> 
> Thinking about it again with your current structure in mind, i now think we should add this option to make it a configurable option

That would be helpful! From a user-perspective I was expecting the .tf files to be picked up in a manner similar to this globstar simulator (since ya'll use the doublestar lib) https://globster.xyz/

PR #1244, opened 2024-03-04
Fix aws-assume-role test cases

ri-roee:
Was running tests for my own development of #1225 and noticed that tests were failing. This PR fixes that by correctly assigning state arns


motatoes:
some tests are not running because we don't have the events in the workflow, going to close the PR and reopen to trigger all tests

motatoes:
oh nice catch 😄 strange that nobody else noticed this until now

Issue #1117, opened 2024-02-05
PR comment Job status - wrong URL to Github Action workflow run


chvima:
Every "Job status" comment contains a link. I would expect this link to redirect to Github Action run, but instead it redirects to URL of the PR itself (which is useless, since you are seeing the comment in that PR).

Example:

👷 Jobs status:
❌ JOB_NAME [failed](https://github.com/OWNER/REPO/pull/PR_ID#)

Expectation:

`https://github.com/OWNER/REPO/actions/runs/RUN_ID`

PR #1267, opened 2024-03-12
Add helm release github action

ri-roee:
There are pre-requisites which I don't have the required permissions to manage, see more here: https://github.com/helm/chart-releaser-action?tab=readme-ov-file#pre-requisites. cc: @motatoes 

This PR enables the use of the new helm chart without copying it locally.

### DIFF
--- a/.github/workflows/helm_release.yml
+++ b/.github/workflows/helm_release.yml
@@ -1,0 +1,32 @@
+name: Release Charts
+
+on:
+  push:
+    branches:
+      - develop
+
+jobs:
+  release:
+    permissions:
+      contents: write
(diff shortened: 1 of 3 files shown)


ri-roee:
Link issue: https://github.com/diggerhq/digger/issues/1117

motatoes:
@ri-roee thanks! Maybe we can split out the PR to fix workflow job url from the helmchart release action? :) that would be great to review them both

Also I noticed tests are failing, not sure why (didnt dig yet) but that's probably because of the second change, unrelated to helmchart

ri-roee:
Splitting out into two PRs

PR #1270, opened 2024-03-13
Fix hyperlink issue in comments

ri-roee:
Issue: https://github.com/diggerhq/digger/issues/1117

Tested using local tests & github [API docs](https://docs.github.com/en/search-github/getting-started-with-searching-on-github/understanding-the-search-syntax#query-for-dates)


motatoes:
nice catch! Didn't notice that haha and have always delayed looking into this. huge win if this works!!

motatoes (approved):

PR #1271, opened 2024-03-14
Make specifying both assume roles optional

ri-roee:
My use-case here is that both the `state` and `command` roles are equivalent and therefore duplicates code so this introduces a cleaner approach.


motatoes (approved):

motatoes:
Cool! I also should mention that we have a new argument called [`workflow_file`](https://github.com/diggerhq/digger/pull/1232) under each project so you can trigger a different workflow file per project. Yes it leads to alot of duplication in github action code but I personally prefer using the assume role in the workflow file as opposed to specifying roles in digger.yml, but that is just me

PR #1277, opened 2024-03-18
Add helm chart lint/test job on KinD

ri-roee:
Grabbed the helm chart lint/test action from [Helm documentation](https://helm.sh/docs/howto/chart_releaser_action/) and made the release conditional on its success. Also tweaked the `default` branch since the actions assume `master` but we use `develop` here.

@motatoes for the release to fully work we need to setup a Github page for this repo (described in more [detail here](https://github.com/helm/chart-releaser-action?tab=readme-ov-file#pre-requisites)). Could you set that up, I don't have the required permissions on this repo :)

### DIFF
--- a/.github/workflows/helm_release.yml
+++ b/.github/workflows/helm_release.yml
@@ -6,7 +6,50 @@ on:
       - develop
 
 jobs:
+  lint-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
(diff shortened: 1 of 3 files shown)


ri-roee:
> Hey @ri-roee I've created a gh-pages branch and its deployed as a page [here](https://diggerhq.github.io/digger/)
> 
> I'm unsure why the action is detecting "no changes" even though all looks good to me, must be something I'm missing

Oh I totally missed that, thanks for taking care of it! Yeah not 100% sure about the no changes part, I bumped the minor version in this PR hopefully it picks that up

Issue #811, opened 2023-11-23
Caching of terraform dependencies


ZIJ:
Digger feels slow; Francoi's suggestion is that we might not be using Github cache for terraform dependencies, so it always runs init from scratch

JoshuaJackson-jobvite:
assuming that digger is running terraform relatively standard, having a .terraformrc that has something like:
```
plugin_cache_dir   = "$HOME/.terraform.d/plugin-cache"
```
and then you use the cache restore/cache before/after digger runs in the pipeline should just work.

JoshuaJackson-jobvite:
Something like the following should work as an example:

```
name: Digger PR flow

on:
  pull_request:
    branches: ["main"]
    types: [closed, opened, synchronize, reopened]
  issue_comment:
    types: [created]
    if: startsWith(github.event.comment.body, 'digger')
  workflow_dispatch:

jobs:
  digger-job:
    runs-on: ubuntu-latest
    permissions:
      contents: write # required to merge PRs
      id-token: write # required for workload-identity-federation
      pull-requests: write # required to post PR comments
      statuses: write # required to validate combined PR status
      packages: read # required to read the packages
      actions: read
    steps:
      # create a temp. token from the github app installed in the org.
      # The app needs to have read access to all private terraform repos  in the org
      # this token is used in the next step to allow cloning terraform module repos
      - uses: actions/cache/restore@v4
        id: restore_cache
        name: restore_cache
        with:
          path: ${{ github.workspace }}/cache
          key: ${{ runner.os }}-cache-${{ hashFiles('**/cache') }}
          restore-keys: |
            ${{ runner.os }}-cache-${{ hashFiles('**/cache') }}
            ${{ runner.os }}-cache
      - uses: diggerhq/digger@v0.4.2
        with:
          setup-aws: false
          no-backend: true
          disable-locking: true
          digger-hostname: "https://cloud.digger.dev"
          digger-organisation: "XXXX"
          digger-token: ${{ secrets.DIGGER_TOKEN }}
        env:
          GITHUB_OWNER: XXXXX
          TF_PLUGIN_CACHE_DIR: ${{ github.workspace }}/cache
      - uses: actions/cache/save@v4
        name: cache-save
        with:
          path: ${{ github.workspace }}/cache
          key: ${{ runner.os }}-${{ hashFiles('**/cache') }}
```

JoshuaJackson-jobvite:
Assuming acceptable I can look into updating the documentation around it as applicable.

[...]

PR #1292, opened 2024-03-20
Feat: Add caching support for dependencies

JoshuaJackson-jobvite:
Why do we need this change?
=======================
When running across multiple projects with github actions, the current behaviors are downloading the plugins each time for each project. This pr adds in a behavior to add a caching directory for the terraform and opentofu related behaviors. This should result in a improvement where multiple projects have the same dependency. Between run's as well we'll restore a cache and should prevent the needs to download again across multiple behaviors.

There's a few catches in this initial mvp pr. There's no cleanup of old version of plugins from the cache so it can grow. Multiple executions could fail to write to the cache if they result in the same hash, but as potential issues come up we can address them separately

This work is based on request in #811 

=======================
* Creates a cache dir
* Adds in support for creating and restoring caches using action/cache
* Sets environment variables for digger's execution to pass to the terraform/opentofu setup to leverage the single cache dir for all plugin needs.
*


action.yml
```diff
@@ -315,6 +334,12 @@ runs:
         PATH=$PATH:$(pwd)
         cd $GITHUB_WORKSPACE
         digger
+
+    - uses: actions/cache/save@v4
+      name: cache-save
+      with:
+        path: ${{ github.workspace }}/cache
+        key: ${{ runner.os }}-${{ hashFiles('**/cache') }}
```
motatoes: should we also have:

```
 if: ${{ inputs.cache-dependencies == 'true' }}
```

here?

eco-JoshuaJackson: Yep realized that when I was reviewing in a break between, so added :)

motatoes:
We have an issue with our other PR event listerns I need to fix, going to close and re-open the PR to trigger rest of tests

vamsi10g (approved):

raghavendraDBA (approved):

action.yml
```diff
@@ -137,6 +137,10 @@ inputs:
     description: 'drift-detection slack drift url'
     required: false
     default: ''
+  cache-dependencies:
+    description: 'Leverage actions/cache to cache dependencies to speed up execution"
```
motatoes: just got to testing this: you have a mismatched quotation mark here:

```suggestion
    description: "Leverage actions/cache to cache dependencies to speed up execution"
```

motatoes (changes requested):
left you a comment regarding mismatched quote to update

action.yml
```diff
@@ -262,6 +277,9 @@ runs:
           core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env['ACTIONS_RUNTIME_TOKEN'])
           core.exportVariable('ACTIONS_RUNTIME_URL', process.env['ACTIONS_RUNTIME_URL'])
 
+    - name: create cache dir
+      run: |
+        mkdir -p $GITHUB_WORKSPACE/cache
```
motatoes, line 281: Need to also add shell: bash because this is failing too
```suggestion
    - name: create cache dir
      run: |
        mkdir -p $GITHUB_WORKSPACE/cache
      shell: bash
```

motatoes:
So my initial basic test in this repo was not successful: https://github.com/motatoes/test-digger-caching

The first job managed to save the cache successfully. However the subsequent job failed to save the cache with already exists error: https://github.com/motatoes/test-digger-caching/actions/runs/8789860900/job/24120563671

I wonder if we are missing a flag to override the cache?

action.yml
```diff
@@ -212,6 +216,17 @@ runs:
...
 
+    - uses: actions/cache/restore@v4
+      id: restore_cache
+      name: restore_cache
+      with:
+        path: ${{ github.workspace }}/cache
+        key: ${{ runner.os }}-cache-${{ hashFiles('**/cache') }}
+        restore-keys: |
+          ${{ runner.os }}-cache-${{ hashFiles('**/cache') }}
+          ${{ runner.os }}-cache
```
motatoes: So if I understand it correctly, at this point of invoking digger there is no cache directory available so whats the point of trying to do  `${{ runner.os }}-cache-${{ hashFiles('**/cache') }}`? why not just `${{ runner.os }}-cache`

Also I think we can prefix it with digger instead of runner.os:

```suggestion
      with:
        path: ${{ github.workspace }}/cache
        key: digger-cache
        restore-keys: |
          digger-cache
```

Which will esentially match the most recent cache

action.yml
```diff
@@ -315,6 +335,14 @@ runs:
         PATH=$PATH:$(pwd)
         cd $GITHUB_WORKSPACE
         digger
+
+    - uses: actions/cache/save@v4
+      name: cache-save
+      with:
+        path: ${{ github.workspace }}/cache
+        key: ${{ runner.os }}-${{ hashFiles('**/cache') }}
```
motatoes: Over here we should add the hash, and also I suggest we use digger-xxxx as a convention

```suggestion
        path: ${{ github.workspace }}/cache
        key: digger-${{ hashFiles('**/cache') }}
```

motatoes (changes requested):
Starting to follow along now, left you a last set of comments and if I understand it correctly can ship it

action.yml
```diff
@@ -315,6 +334,14 @@ runs:
         PATH=$PATH:$(pwd)
         cd $GITHUB_WORKSPACE
         digger
+
+    - uses: actions/cache/save@v4
+      name: cache-save
+      with:
+        path: ${{ github.workspace }}/cache
+        key: digger-${{ hashFiles('**/cache') }}
```
JoshuaJackson-jobvite, line 342:
```suggestion
        key: digger-cache-${{ hashFiles('**/cache') }}
```

Think we might need it this way, as digger-${{ hashFiles }} would not be a match for digger-cache in any way, but digger-cache-abc123 should be a general match

motatoes (changes requested):
just one last thing to change "digger-${{ hashFiles('**/cache') }}" -> digger--cache-${{ hashFiles('**/cache') }}

action.yml
eco-JoshuaJackson:
```suggestion
        key:  digger-cache-${{ hashFiles('**/cache') }}
```

action.yml
```diff
@@ -315,6 +334,14 @@ runs:
         PATH=$PATH:$(pwd)
         cd $GITHUB_WORKSPACE
         digger
+
+    - uses: actions/cache/save@v4
+      name: cache-save
+      with:
+        path: ${{ github.workspace }}/cache
+        key:  digger-cache-${{ hashFiles('**/cache') }}
```
eco-JoshuaJackson:
```suggestion
        key: digger-cache-${{ hashFiles('**/cache') }}
```

motatoes (approved):

PR #1294, opened 2024-03-20
Attempt fix for AWS provider assumerole in orchestration workflow

ri-roee:


libs/orchestrator/json_models.go
```diff
@@ -12,19 +16,21 @@ type StageJson struct {
...
+	Commands           []string                          `json:"commands"`
+	ApplyStage         StageJson                         `json:"applyStage"`
+	PlanStage          StageJson                         `json:"planStage"`
+	PullRequestNumber  *int                              `json:"pullRequestNumber"`
+	EventName          string                            `json:"eventName"`
+	RequestedBy        string                            `json:"requestedBy"`
+	Namespace          string                            `json:"namespace"`
+	StateEnvVars       map[string]string                 `json:"stateEnvVars"`
+	CommandEnvVars     map[string]string                 `json:"commandEnvVars"`
+	StateEnvProvider   *stscreds.WebIdentityRoleProvider `json:"stateEnvProvider"`
```
motatoes: This is serialization at the point of orchestrator, we won't be able to serialize providers back to the cli, so we should only store the roles as strings

StateEnvRole string
CommandEnvRole string

libs/orchestrator/json_models.go
```diff
@@ -37,37 +43,41 @@ func (j *JobJson) IsApply() bool {
...
+		Commands:           job.Commands,
+		ApplyStage:         stageToJson(job.ApplyStage),
+		PlanStage:          stageToJson(job.PlanStage),
+		PullRequestNumber:  job.PullRequestNumber,
+		EventName:          job.EventName,
+		RequestedBy:        job.RequestedBy,
+		Namespace:          job.Namespace,
+		StateEnvVars:       job.StateEnvVars,
+		CommandEnvVars:     job.CommandEnvVars,
+		StateEnvProvider:   job.StateEnvProvider,
```
motatoes: So by the time we get here we will have StateEnvRole and CommandEnvRole as strings and we need to convert them to providers using `orchestrator.GetProviderFromRole()` but also check for "" in that function and return nil if so

motatoes (changes requested):

libs/orchestrator/json_models.go
ri-roee: Slightly confused about this point. We're storing them as strings but the return value of the `orchestrator.GetProviderFromRole()` is a `*stscreds.WebIdentityRoleProvider`?

PR #1296, opened 2024-03-20
Fix: ensure that variables used for slack match the input variable name

### LINKED ISSUE #1295: Actions: Drift detection url for slack is different in input selection from variable name
In the main.go we see the following lookup of env variable:
```
			slackNotificationUrl := os.Getenv("INPUT_DRIFT_DETECTION_SLACK_NOTIFICATION_URL")
```

in the action.yaml for either version of the execution we do:

```
        INPUT_DRIFT_DETECTION_SLACK_NOTIFICATION_URL: ${{ inputs.drift-detection-slack-drift-url }}
```

So the env variable we set is correct, and we see that we are looking at inputs: drift-detection-slack-drift-url 

variable in the action however is https://github.com/diggerhq/digger/blob/4a6573a508ee482086182fad6562b2c6d7573978/action.yml#L136-L139
```
  drift-detection-slack-notification-url:
```

which is not the same as what the inputs search is. Thus we break the behaviors and its always a null.

JoshuaJackson-jobvite:
Why do we need this change?
=======================
Introduced in #1239 a change went into the action to change the inputs for the digger portion of the action but didn't update the inputs to match. So this change ensures that inputs and usage are consistent. This prevents the slack_url from being a NULL always and causing drift detection to always fail out.

=======================
* ensure that slack url variables are consistent.
* Fixes #1295

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -279,7 +279,7 @@ runs:
         INPUT_DIGGER_PROJECT: ${{ inputs.project }}
         INPUT_DIGGER_MODE: ${{ inputs.mode }}
         INPUT_DIGGER_COMMAND: ${{ inputs.command }}
-        INPUT_DRIFT_DETECTION_SLACK_NOTIFICATION_URL: ${{ inputs.drift-detection-slack-drift-url }}
+        INPUT_DRIFT_DETECTION_SLACK_NOTIFICATION_URL: ${{ inputs.drift-detection-slack-notification-url }}.
         NO_BACKEND: ${{ inputs.no-backend }}
         DEBUG: 'true'
       run: |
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -279,7 +279,7 @@ runs:
         INPUT_DIGGER_PROJECT: ${{ inputs.project }}
         INPUT_DIGGER_MODE: ${{ inputs.mode }}
         INPUT_DIGGER_COMMAND: ${{ inputs.command }}
-        INPUT_DRIFT_DETECTION_SLACK_NOTIFICATION_URL: ${{ inputs.drift-detection-slack-drift-url }}
+        INPUT_DRIFT_DETECTION_SLACK_NOTIFICATION_URL: ${{ inputs.drift-detection-slack-notification-url }}.
```
ri-roee: Small change :) 

```suggestion
        INPUT_DRIFT_DETECTION_SLACK_NOTIFICATION_URL: ${{ inputs.drift-detection-slack-notification-url }}
```

ri-roee (approved):

motatoes (approved):

PR #1297, opened 2024-03-20
Correctly pass project roles from orchestrator to cli

motatoes:
We have been not passing digger.yml project-level roles from backend to cli and therefore they were not working properly with orchestrator triggered jobs. We are now passing them and during unmarshal of the serialized json spec we set the right env provider in the job struct


ri-roee (approved):
blindly giving a LGTM 👍🏼

UtpalJayNadiger (approved):

PR #1312, opened 2024-03-28
Fix GH comment hyperlink

ri-roee:
I found two bugs with local testing:

1) The `-07:00` UTC offset portion of the timeFilter was throwing off the github query so I got rid of it alltogether since I'm assuming the API is normalized relative to the query time anyways

2) The code is using the comparison`strings.Contains(*step.Name, diggerJobID)` however the id was never set as part of the step name. I believe this is because the actions doesn't take it in as an input, despite the action attempting to use it and echo it.

I'd like to push this to a test release and do further testing since it was hard to write mock tests for.


ri-roee:
All that's required is the following step!

```
      - name: ${{github.event.inputs.id}}
        run: echo job id ${{ inputs.id }}   
```

PR #1315, opened 2024-03-29
Fix: Enable per project assumerole support in drift detection

### LINKED ISSUE #1133: Drift-detection mode isn't paying attention to the `aws_role_to_assume` config
User Feedback:

>Running in drift-detection mode does not assume an AWS role so drift-detection is failing for me currently.

From user K.B

JoshuaJackson-jobvite:
Why do we need this change?
=======================
Drift detection currently does not take the properties from the digger.yml and do authentication changes to ensure that it can review the live resources for a given account. This attempts to fix that behavior to allow drift detection to work

=======================
* Add StateEnvProvider and CommandEnvProvider which are used in PopulateAwsCredentialsEnvVarsForJob to gather and do the appropriate adjustments.

should fix #1133


JoshuaJackson-jobvite:
Apologies for the multiple commits related to others I'd not planned on multiple fixes being needed and my develop still has changes that I don't want to close/reopen the pr related to. So will try and keep this clean

motatoes (approved):

Issue #1261, opened 2024-03-08
TF Plan security: Full plan in output when checking for plan violation


chvima:
v0.4.6

After I've implemented Plan policies in our projects, the output in Github Actions contains full terraform plan (in the format of input to plan policy OPA check)

Example:
```

Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # null_resource.echo will be created
  + resource "null_resource" "echo" {
      + id = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.

─────────────────────────────────────────────────────────────────────────────

Saved the plan to:
...

To perform exactly these actions, run the following command to apply:
    terraform apply "..."
Running command: terraform [show -no-color -json ...]
no $DIGGER_OUT file at: /__w/_temp/digger-out.log
DEBUG: passing the following input policy: map[terraform:map[configuration:map[provider_config:map[aws:map[expressions:map[default_tags:[map[tags:map[references:[module.tags.tags module.tags]]]] profile:map[constant_value:mosaic-dev] region:map[constant_value:eu-west-1]] full_name:registry.terraform.io/hashicorp/aws name:aws version_constraint:~> 5.0] module.tags.module.backend_config:terraform:map[full_name:terraform.io/builtin/terraform module_address:module.tags.module.backend_config name:terraform] module.tags:external:map[full_name:registry.terraform.io/hashicorp/external module_address:module.tags name:external] null:map[full_name:registry.terraform.io/hashicorp/null name:null] terraform:map[full_name:terraform.io/builtin/terraform name:terraform]] 
... FULL PLAN IS HERE
```

This can cause security issues as all users with read access to the repo can see this and TF plan is generally considered to be sensitive and stored in encrypted locations.

chvima:
Fixed in https://github.com/diggerhq/digger/pull/1373 release in https://github.com/diggerhq/digger/releases/tag/v0.4.16

PR #1373, opened 2024-04-25
Hotfix-Disable logging of full Terraform Plan output in CheckPlanPolicy

chvima:
This is a hotfix for my issue https://github.com/diggerhq/digger/issues/1261

The "input" variable that is removed from logs in this PR contained full output of terraform plan. The output from TF plan can be sensitive (as it can contain secrets if those are managed by Terraform) and logging it has caused it to be public in the output of  Github Action (those are public to all users with read access to the repository)

Proper fix would be implementing a better logging in the Digger CLI application, with log levels.
Currently, the log message is marked as DEBUG, but there is no notion of log levels implemented in Digger CLI application and those "DEBUG" logs are contained in standard output for every run. 

But due to the high impact of this issue on security I would be glad if you could implement this in nearest release.

### DIFF
--- a/cli/pkg/policy/policy.go
+++ b/cli/pkg/policy/policy.go
@@ -352,7 +352,7 @@ func (p DiggerPolicyChecker) CheckPlanPolicy(SCMrepository string, SCMOrganisati
 	}
 
 	ctx := context.Background()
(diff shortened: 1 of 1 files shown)


motatoes (approved):
Thanks for the contribution!

PR #1403, opened 2024-04-27
use dynamodb consistent read to ensure we check against latest lock

larhauga:
I am not super familiar with ConsistentRead, but it seems like a good option to use when utilizing dynamodb as a lock provider. Ref https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadConsistency.html


motatoes:
Nice, will definitely help for companies with high volume of PRs 😉 
BTW we are gonna be moving to locks being in the orchestrator backend since it will help improve alot of experience around the locking part. So the cloud native locks will probably be only applicable to folks using backendless mode

motatoes (approved):

PR #1429, opened 2024-05-06
Updated the docs related to storing plans in AWS

ben-of-codecraft:
This updates some missing documentation how to use a different place for Upload destination, specifically S3.  I move the article out of GCP since it is no longer GCP specific.


motatoes:
Thanks!

PR #1466, opened 2024-05-15
Added bool that was missing for LoadDiggerConfig

ben-of-codecraft:
I was getting the following error when trying to build dgctl to test the new cli

```sh
cmd/validate.go:21:59: not enough arguments in call to digger_config.LoadDiggerConfig
	have (string)
	want (string, bool)
```

It looks like the arg is to generate projects, so made the assumption this should be false for validation of config.

### DIFF
--- a/dgctl/cmd/validate.go
+++ b/dgctl/cmd/validate.go
@@ -6,10 +6,11 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


motatoes:
Oh yeah! I think we want to be "true" though because it will aid in testing out the digger.yml how it generates the projects for a certain directory structure, very useful for debugging! :D

motatoes (approved):

Issue #58, opened 2023-03-23
Show compact diff


ZIJ:
"if I apply a one line change in big YAML for helm_release, in the output I'll see two gigantic documents with one different line which makes the output near unreadable.

I think, if you will able to make compact but informative output it may be a big plus."

by @AntonChekunov

dangarthwaite:
I had the same thought.  https://github.com/dineshba/tf-summarize is what you want, here.

PR #1497, opened 2024-05-23
tf-summarize to comment on plan

larhauga:
This pull requests add a summary to the comment using tf-summarize table style.
We think that this will reduce the changes of introducing errors as it gives a easy accessible and comprehensive view of the changes to be made.

Also mentioned in https://github.com/diggerhq/digger/issues/58

image

Do we need a toggle to enable this, or should it be enabled by default this PR does?

### DIFF
--- a/libs/terraform_utils/plan_summary.go
+++ b/libs/terraform_utils/plan_summary.go
@@ -1,10 +1,16 @@
 package terraform_utils
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/samber/lo"
 	"sort"
+
+	"github.com/dineshba/tf-summarize/terraformstate"
+	"github.com/dineshba/tf-summarize/writer"
+	tfjson "github.com/hashicorp/terraform-json"
+
+	"github.com/samber/lo"
 )
 
 type PlanSummary struct {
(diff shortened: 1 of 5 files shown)


motatoes:
Lovely! 🥇  Just showed your screenshot to the team and we all got really excited by this change! We think it would be great to show it by default since its much nicer than looking at raw plans for sure. Thanks for this valuable contribution!

libs/terraform_utils/plan_summary_test.go
```diff
@@ -50,3 +51,10 @@ func TestPlanFootprintSimilarity(t *testing.T) {
...
+	isEmpty, err := GetTfSummarizePlan(nonEmptyTerraformPlanJson)
+	assert.Nil(t, err)
+	assert.NotEmpty(t, isEmpty)
```
motatoes: small nit about the variable name `isEmpty` we should probably rename it to planSummary to make the test more readable

motatoes (changes requested):

libs/terraform_utils/plan_summary.go
```diff
@@ -132,3 +138,17 @@ func SimilarityCheck(footprints []TerraformPlanFootprint) (bool, error) {
 	return allSimilar, nil
 
 }
+
+func GetTfSummarizePlan(planJson string) (string, error) {
+	plan := tfjson.Plan{}
```
motatoes, line 143: wow TIL that we can rely on the types in the library rather than bake our own types as in this file!

libs/terraform_utils/plan_summary_test.go
larhauga: Makes sense :+1: :white_check_mark:

libs/terraform_utils/plan_summary.go
larhauga: It's pretty neat :) I think that OpenTofu still uses `github.com/hashicorp/terraform-json`, so it looks viable to continue using it when having both terraform and OpenTofu in mind.

motatoes (approved):

Issue #1535, opened 2024-05-30
Golang caching does not cache dependencies on action builds


ben-of-codecraft:
If not using a versioned request Golang will not cache. This creates a longer builds as dependencies have to be downloaded every time.  This is a bug with how composite actions and setup-go work together that has been filed. 

I found an alternate action that does effectively the same thing without using the built-in setup-go caching. 
https://github.com/magnetikonline/action-golang-cache/blob/main/action.yaml

Unexpected Result: 
![image](https://github.com/diggerhq/digger/assets/110695027/bb43cef3-7080-49ad-a47b-8ae48f52e3de)

Updated action code: 
```yaml
    - name: Setup OpenTofu
      uses: opentofu/setup-opentofu@v1.0.3
      with:
        tofu_version: ${{ inputs.opentofu-version }}
      if: inputs.setup-opentofu == 'true'


    - name: Setup Checkov
      run: |
        python3 -m venv .venv
        source .venv/bin/activate
        pip3 install --upgrade pip
        pip3 install --upgrade setuptools
        pip3 install -U checkov==${{ inputs.checkov-version }}
      shell: bash
      if: inputs.setup-checkov == 'true'

    - name: Setup go with cache  
      uses: magnetikonline/action-golang-cache@v5
      with:
        go-version-file: ${{ github.action_path }}/cli/go.mod          
      if: ${{ !startsWith(github.action_ref, 'v') && github.action_ref != 'latest'}}
  ```
  
  With the changes above result: 
  
![image](https://github.com/diggerhq/digger/assets/110695027/b8f896f1-2a2a-42ef-8773-bac326b9b356)

**Overall this took repeat builds from ~2-3 min down to 30-40s**

motatoes:
Hey! Thanks for filing this. For everyone's context we currently have to ways of consuming the digger action. The first is through downloading prebuilt binaries and invoking `digger`. The second is to build from source and its useful for testing from a branch that you are actively working on (development). 

We build digger binaries on release and we have a convention that release names start with "v". So that is what we check for in the action as well. If it starts with a "v" then we download the available binary, otherwise we clone the repo and build from source. 

If you wish to use digger with always latest we have a release called "vLatest" which is always a release pointing to the latest release branch ..

[...]

PR #1573, opened 2024-06-10
Cache Go dependencies when building CLI

sikha-root:
Enables dependency caching of the Digger CLI binary being built in a deploy when not using version tags (I'm using hashes for immutability). Since it seems I can't use `hashFiles` for files outside of the workspace of the Terraform code repo running the action we copy that file to one that shouldn't collide with any source files so that we can ensure the cache key is sufficiently unique (maybe not unique enough? not sure).

I'm reusing the `cache-dependencies` input here because of the name and the description provided indicating that it should probably cover this instead of just Terraform providers, and people turning that on likely want this too.

Also pins the go version used by setup-go to the one in `cli`'s dependency file.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -264,7 +264,6 @@ runs:
         tofu_wrapper: false
       if: inputs.setup-opentofu == 'true'
 
-
     - name: Setup Checkov
       run: |
         python3 -m venv .venv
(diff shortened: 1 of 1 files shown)


sikha-root:
Finally working at least in a way that's usable. This should close #1535 unless something different is desired.

sikha-root:
I'm assuming the CI failure here and in my other PR are just a result of external PRs not having access to a repository secret or something. Should be resolved separately I guess.

motatoes (approved):
Thanks so much for this!! Going to give it a test drive

Issue #1577, opened 2024-06-13
Proposal: move disable-locking flag to digger.yml


motatoes:
Currently we have disable locking in [action.yml](https://github.com/diggerhq/digger/blob/develop/action.yml#L105) and it was working fine for both backendless and orchestrator modes.

However we recently moved orchestrator locking to be in database for performance reasons (https://github.com/diggerhq/digger/pull/1524) .

Now we have introduced another issue because there is no longer a way to disable locking if you are using digger with orchestrator mode (the flag was only avaiable in cli and since with orchestrator locking no longer performed in the cli it means that you can no longer disable locking by passing `disable-locking` to the action directly.

Propose to move that flag to a top level digger.yml value since then both orchestrator and cli can read it unified and have consistent behaviour:

```
# digger.yml

disable_locking: true

projects:
    - name: balbla
      dir: bla
```

Then the backend can pick this up in the webhook and if set then we don't perform any locking checks. Also we deprecate the older disable-locking from the action and rely on digger.yml to perform this locking disabling moving forward

motatoes:
This is now done using a flag called `pr_locks: false` instead of `disable_locking: true`

https://docs.digger.dev/howto/disable-locking

PR #1578, opened 2024-06-14
Add PR locking configuration at the digger.yml level

evanstachowiak:
Following the conversation here https://diggertalk.slack.com/archives/C01DVJRK6CX/p1718200181031169.

Adding the ability to enable/disable PR locking in GitHub with a digger.yml config setting.

Addresses this issue: https://github.com/diggerhq/digger/issues/1577

### DIFF
--- a/backend/controllers/github.go
+++ b/backend/controllers/github.go
@@ -486,9 +486,10 @@ func handlePullRequestEvent(gh utils.GithubClientProvider, payload *github.PullR
 	}
 
 	// perform locking/unlocking in backend
-	if dg_configuration.PrLocks{
+	if config.PrLocks {
 		for _, project := range impactedProjects {
 			prLock := dg_locking.PullRequestLock{
+				Enable: config.PrLocks,
 				InternalLock: locking.BackendDBLock{
 					OrgId: organisationId,
 				},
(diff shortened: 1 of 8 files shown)


backend/controllers/github.go
```diff
@@ -486,21 +486,23 @@ func handlePullRequestEvent(gh utils.GithubClientProvider, payload *github.PullR
...
-			Reporter:         comment_updater.NoopReporter{},
-			ProjectName:      project.Name,
-			ProjectNamespace: repoFullName,
-			PrNumber:         prNumber,
-		}
-		err = PerformLockingActionFromCommand(prLock, *diggerCommand)
-		if err != nil {
-			utils.InitCommentReporter(ghService, prNumber, fmt.Sprintf(":x: Failed perform lock action on project: %v %v", project.Name, err))
-			return fmt.Errorf("failed to perform lock action on project: %v, %v", project.Name, err)
+	if dg_configuration.PrLocks{
```
motatoes: you need to use the config object here:

```suggestion
	if config.PrLocks{
```

motatoes: you need to do the same if statement in the `handleIssueComment` webhook handler in the same file

motatoes (commented):
Thanks for this. Overall looks good and I'm happy about the naming. 

We also need to override locking to be noopLock if this flag is set for backendless mode here: https://github.com/diggerhq/digger/blob/1d611b44e0af3efbb841c22fdd305a086f2474f7/cli/pkg/github/github.go#L214

evanstachowiak:
I added support for `Enable` on the struct level, I couldn't see a better way to do it for the cli. In this case I could probably also remove these `if` statements around the `for` loops, but it would also save some processing to leave them in.

motatoes:
Hey @evanstachowiak this looks great, thanks! I would rather not have a flag on the PrLock interface. We have a NoopLock interface implemented here: https://github.com/diggerhq/digger/blob/e43db334eaefb55c592bca7cc98d57fd524b74ca/libs/locking/locking.go#L36 which will ignore any lock operations

so if you had something like 

```
if config.Prlocks {
    log.Printf("info: using NoopLock)
    lock = NoopLock{}
}
```

[here](https://github.com/diggerhq/digger/blob/1d611b44e0af3efbb841c22fdd305a086f2474f7/cli/pkg/github/github.go#L214) it will do the trick - since that path is outside of the orchestrator path.

motatoes:
@evanstachowiak actually thinking about it again, your changes will work well, the enable flag works. Future refactoring can be done but this is good enough for now!

motatoes (approved):

PR #1579, opened 2024-06-14
This adds PR comments with digger run fails 

ben-of-codecraft:
When a digger job would fail, the error message in the log did not show the output, it only reported it to the backend API.  This information is helpful in debugging problems.   

Additionally, this adds a PR comment with the error message.

### DIFF
--- a/cli/pkg/digger/digger.go
+++ b/cli/pkg/digger/digger.go
@@ -3,15 +3,16 @@ package digger
 import (
 	"errors"
 	"fmt"
-	coreutils "github.com/diggerhq/digger/libs/comment_utils/utils"
-	locking2 "github.com/diggerhq/digger/libs/locking"
 	"log"
 	"os"
 	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ben-of-codecraft:
This is what is failing 

```
    license_test.go:10: 
        	Error Trace:	/home/runner/work/digger/digger/libs/license/license_test.go:10
        	Error:      	Received unexpected error:
        	            	license key is not valid: {"valid":false}
        	Test:       	TestLicenseKeyChecker
--- FAIL: TestLicenseKeyChecker (0.93s)
FAIL
FAIL	github.com/diggerhq/digger/libs/license	0.932s
```

Not sure why this one is failing any thoughts?

motatoes:
This looks great, but we prefer to use `reporter` object when reporting since it will append it to the same comment for the job isntead of creating a new comment. See example usage here: https://github.com/diggerhq/digger/blob/3496e66ec5591dce1c4b139323f4a9c39a2c53c0/cli/pkg/digger/digger.go#L321

PR #1597, opened 2024-07-01
Feat/use digger spec

motatoes:
- This is a breaking change for github orchestrator to introduce digger spec in plance of "job"
- bckendless mode is not impcted by this change
- We explain about digger spec here: https://docs.digger.dev/securing-digger/spec-signing
- Users updating to the next version will need to update their digger_workflow.yml to only accept a `spec` (no more job, comment_id, id) and pass it on to digger action as digger-spec argument:

```
name: Digger Workflow

on:
  workflow_dispatch:
    inputs:
      id:
        description: 'run identifier'
        required: false
      spec:
        required: true

......
.......


    steps:
      - name: digger run
        uses: diggerhq/digger@vLatest
        with:
          digger-spec: ${{ inputs.spec }}

```

### DIFF
--- a/docs/cloud-providers/authenticating-with-oidc-on-aws.mdx
+++ b/docs/cloud-providers/authenticating-with-oidc-on-aws.mdx
@@ -13,12 +13,8 @@ on:
       id:
         description: 'run identifier'
         required: false
-      job:
+      spec:
         required: true
-      comment_id:
-        required: true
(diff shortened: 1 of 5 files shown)


motatoes:
digger plan

UtpalJayNadiger (approved):

Issue #1601, opened 2024-07-02
github action and aws documentation missing a few indents for working configuration


JoshuaJackson-jobvite:
While looking at upgrading to the newest version noticed that in most places its correct for the documentation, but for aws and github actions it had inputs at the wrong level of indentation.

PR #1602, opened 2024-07-02
Fix: Correct indentation on github-actions with aws inputs

JoshuaJackson-jobvite:
Why do we need this change?
=======================
The live documentation had inputs at the same level as the workflow-dispatch, this will result in an error if folks just copy/paste the results and attempt to run.

=======================
* Fixes the indentation that the inputs are required in the workflow_dispatch:, and not at the same level.
* Fixes: #1601 1601


motatoes:
Thanks!

PR #1671, opened 2024-08-19
feat/plan summary in collapsible

motatoes:
Make the plan summary in its own collapsible comment and collapsed by default. Before:

Screenshot 2024-08-19 at 16 26 25

After (plan summary gets its own collapsible comment and is collapsed by default):

Screenshot 2024-08-19 at 16 26 29


ZIJ (approved):

tanguyantoine:
could this be configured through env variable? I find it useful to have it the old way (not collapsed)

motatoes:
> could this be configured through env variable? I find it useful to have it the old way (not collapsed)

@tanguyantoine just filed a separate issue to track this request :)

PR #1713, opened 2024-09-17
Stop automatically upgrading module/provider versions

kieranbrown:
Please feel free to close this Pull Request if this was an intentional design decision, but I feel as though Digger should not upgrade modules/providers as this can lead to discrepancies when running plans locally vs the CI. Digger should adhere to the project's terraform lock file.

## Release Notes

```
- Digger no longer automatically upgrades modules/providers with the `--upgrade` flag when initializing a project. To restore the old behaviour you can use the `extra_args` field to pass `--upgrade` in the plan/apply steps in digger.yaml (https://docs.digger.dev/ce/reference/digger.yml#example-using-all-keys).
```

### DIFF
--- a/libs/execution/opentofu.go
+++ b/libs/execution/opentofu.go
@@ -16,7 +16,6 @@ type OpenTofu struct {
 }
 
 func (tf OpenTofu) Init(params []string, envs map[string]string) (string, string, error) {
-	params = append(params, "-upgrade=true")
 	params = append(params, "-input=false")
 	params = append(params, "-no-color")
 	stdout, stderr, _, err := tf.runOpentofuCommand("init", true, envs, params...)
--- a/libs/execution/tf.go
+++ b/libs/execution/tf.go
@@ -25,7 +25,6 @@ type Terraform struct {
 }
 
 func (tf Terraform) Init(params []string, envs map[string]string) (string, string, error) {
-	params = append(params, "-upgrade=true")
 	params = append(params, "-input=false")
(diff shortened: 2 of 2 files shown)


motatoes:
Hi @kieranbrown thanks for filing this! I don't have strong opinions about this, Not sure if there's any case fo **having** that flag set by users in which case we should make it considerable. Happy to merge this and add a note in release notes about it if users mention that they prefer it we can bring it back with a flag

kieranbrown:
> Hi @kieranbrown thanks for filing this! I don't have strong opinions about this, Not sure if there's any case fo **having** that flag set by users in which case we should make it considerable. Happy to merge this and add a note in release notes about it if users mention that they prefer it we can bring it back with a flag

Hi @motatoes! Thanks for the fast response. I recently came across an issue where I was seeing a difference in planned results locally vs Digger as a result of an upstream provider change. I wasn't seeing this locally because my provider version was locked. I do feel as though the better default behaviour is to have Digger adhere to the lock file.

Do you require me to make any additional changes to support the note in the release notes?

kieranbrown:
@motatoes I've suggested a release note in the PR description - let me know if you require any more changes for this to be merged.

motatoes:
@kieranbrown many thanks for this, indeed people can add extra args with extra_args, nearly forgot about that 😅

motatoes:
Failing test is due to missing license key secret missing in fork(unrelated issue), happy to merge this now

motatoes (approved):

PR #1720, opened 2024-09-20
Fixes issues with AWSAssumeRole in Blocks for Terraform being passed in

### LINKED ISSUE #1714: Terragrunt - AwsAssumeRole does not work on generate_projects / blocks configuration - Community Edition
### Issue 
When using backendless action inputs:
```yaml
      - name: digger run
        uses: diggerhq/digger@v0.6.39
        with:
            setup-aws: false
            setup-terragrunt: true
            setup-checkov: false
            disable-locking: true
            no-backend: true
            terragrunt-version: 0.50.2
```

The command/state role assumes do not get used correctly when set on the blocks like as follows

```yaml
generate_projects
    blocks:
      - block_name: dev
        terragrunt: true
        root_dir: "dev/"
        workflow: default
        workflow_file: digger_workflow.yml
        aws_role_to_assume:
           aws_role_region: "us-east-1"
           command: "xxxx"
  
```

### Expected Behavior

AWSRole identity provider is used and passed into Terragrunt run similar to to how Terraform works.

ben-of-codecraft:
Fixes #1714

### DIFF
--- a/libs/digger_config/digger_config.go
+++ b/libs/digger_config/digger_config.go
@@ -269,15 +269,23 @@ func HandleYamlProjectGeneration(config *DiggerConfigYaml, terraformDir string, 
 							workflow = b.Workflow
 						}
 
-						err := hydrateDiggerConfigYamlWithTerragrunt(config, TerragruntParsingConfig{
+						tgParsingConfig := TerragruntParsingConfig{
 							CreateProjectName: true,
 							DefaultWorkflow:   workflow,
 							WorkflowFile:      b.WorkflowFile,
 							FilterPath:        path.Join(terraformDir, *b.RootDir),
-						}, terraformDir)
+						};						
+
+						// allow blocks to pass in roles that can be assummed by aws 
+						if b.AwsRoleToAssume != nil {
+							tgParsingConfig.AwsRoleToAssume = b.AwsRoleToAssume
+						}
+
+						err := hydrateDiggerConfigYamlWithTerragrunt(config, tgParsingConfig, terraformDir)
 						if err != nil {
 							return err
 						}
+						
 					}
 				} else {
 					includePatterns = []string{b.Include}
@@ -500,6 +508,7 @@ func hydrateDiggerConfigYamlWithTerragrunt(configYaml *DiggerConfigYaml, parsing
 			WorkflowFile:    &workflowFile,
 			IncludePatterns: atlantisProject.Autoplan.WhenModified,
 			Generated:       true,
+			AwsRoleToAssume: parsingConfig.AwsRoleToAssume,
 		})
 	}
 	return nil
--- a/libs/execution/execution.go
+++ b/libs/execution/execution.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -217,6 +218,7 @@ func (d DiggerExecutor) Plan() (*terraform_utils.TerraformSummary, bool, bool, s
 		}
 	}
 	for _, step := range planSteps {
+		log.Printf(" Running step: %v\n", step.Action)
 		if step.Action == "init" {
 			_, stderr, err := d.TerraformExecutor.Init(step.ExtraArgs, d.StateEnvVars)
 			if err != nil {
@@ -530,4 +532,16 @@ func cleanupTerraformPlan(nonEmptyPlan bool, planError error, stdout string, std
 
 func (d DiggerExecutor) projectId() string {
 	return d.ProjectNamespace + "#" + d.ProjectName
+}
+
+// this will log an exit code and error based on the executor of the executor drivers are by filename
+func logCommandFail(exitCode int, err error) {
+
+	_, filename, _, ok := runtime.Caller(1);
+	if ok {
+		executor := strings.TrimSuffix(path.Base(filename), path.Ext(filename))
+		log.Printf("Command failed in %v with exit code %v and error %v", executor, exitCode, err)
+	} else {
+		log.Printf("Command failed in unknown executor with exit code %v and error %v", exitCode, err)
+	}
 }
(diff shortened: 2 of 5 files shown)


libs/digger_config/digger_config.go
```diff
@@ -269,15 +269,23 @@ func HandleYamlProjectGeneration(config *DiggerConfigYaml, terraformDir string,
...
+						tgParsingConfig := TerragruntParsingConfig{
 							CreateProjectName: true,
 							DefaultWorkflow:   workflow,
 							WorkflowFile:      b.WorkflowFile,
 							FilterPath:        path.Join(terraformDir, *b.RootDir),
-						}, terraformDir)
+						};						
+
+						// allow blocks to pass in roles that can be assummed by aws 
+						if b.AwsRoleToAssume != nil {
```
motatoes: nitpick:  I think its safe to pass b.AwsRoleToAssume directly into the struct as it will be passed into the project as nil and handled well in downstream functions

libs/scheduler/aws.go
```diff
@@ -45,6 +45,10 @@ func (job *Job) PopulateAwsCredentialsEnvVarsForJob() error {
 		log.Printf("Project-level AWS role detected, Assuming role for project: %v", job.ProjectName)
 		var err error
 		backendConfigArgs, err := populateretrieveBackendConfigArgs(*job.StateEnvProvider)
+
+		// Terragrunt will cause a backend configuration problem if backend-config options are passed and envs of the same key are passed. 
+		// which will trigger a request to init with --reconfigure, so do not use backend-config for terragrunt	
+		if job.Terragrunt != true {
```
motatoes, line 51: nitpick: the body of this if statement is not indented properly

libs/execution/execution.go
```diff
@@ -531,3 +533,15 @@ func cleanupTerraformPlan(nonEmptyPlan bool, planError error, stdout string, std
...
+func logCommandFail(exitCode int, err error) {
+
+	_, filename, _, ok := runtime.Caller(1);
+	if ok {
+		executor := strings.TrimSuffix(path.Base(filename), path.Ext(filename))
+		log.Printf("Command failed in %v with exit code %v and error %v", executor, exitCode, err)
+	} else {
+		log.Printf("Command failed in unknown executor with exit code %v and error %v", exitCode, err)
+	}
+}
```
motatoes, line 547: thanks so much for this improvement, I remember terragrunt jobs always had obscure error messages especially when they failed

libs/scheduler/aws.go
motatoes, line 51: today I learned something new, so with terragrunt does it pass those options to terraform based on terragrunt.hcl configuration or what is the source of the duplicates?

libs/scheduler/aws.go
```diff
@@ -62,7 +66,13 @@ func (job *Job) PopulateAwsCredentialsEnvVarsForJob() error {
...
 			return fmt.Errorf("Failed to get (state) keys from role: %v", err)
 		}
-
+		} else {
+			job.StateEnvVars, err = populateKeys(job.StateEnvVars, *job.StateEnvProvider)
+			if err != nil {
+				log.Printf("Failed to get keys from role (StateEnvProvider): %v", err)
+				return fmt.Errorf("Failed to get (state) keys from role: %v", err)
+			}
+		}
```
motatoes: So this is not a challenge and I'm probably wrong because I haven't looked at this code in a while but from memory I remmber that the reason for using "backend configuration" passed as arguments rather than env variables was to tell terraform to pull S3 state from an account different from the account where it is creating resources. 

From my basic tests it is not possible to achieve this with env variables during the init phase. SO, I wonder if this entire else block is needed at all. Because if terragrunt has its own way to pass the same config then maybe we ignore the else entirely.

I'm pretty sure I missed something but very useful to discuss this on here for historical context of the code

motatoes (commented):
The PR looks great, just dropped you a few nits and clarifications, other than that I think its ready to go

libs/scheduler/aws.go
ben-of-codecraft: The StateEnvVars only has 3 values from what I can tell: access key, secret and token.  There would not be anything  specific in there related to which account it sets up.  If the backend is specifying a different account then the state role that gets used would have access to that account via trust policy and get the correct  credentials to pass into the `init` environment command.  There may be some where else that is tying that provider to an account I am not sure.  

The environment variables are passed into the subshell command later so command / state env vars can still be different if using two accounts one for state management and one for actual resources.   Thats not how we do it but I can see why some people would set it up that way if they uses remote states across many projects.

libs/scheduler/aws.go
ben-of-codecraft: From the docs I read it should not happen but in practice I tested it out and it consistently happened, if I set aws credentials in the my environment and then also pass in backend-config options for credentials causes that request on state change when it runs the plan. 

The `init` would work fine with both backend-config and environment variables. 
When the `plan` would run, since backend-config variables were passed only into the `init` step it would detect a remote change even though the backend file was the same (must add some hash or something of the complete config in terragrunt / terraform somewhere)

So the option was to make it pass backend-config to all steps or just exclude it.  I chose to exclude it path, as it does not seem to impact terraform. backend-config is not recommended on the TF site, I figured it was best at least for TG to just ignore their use. 

I did not go much deeper than that, but I know the above does work.

ben-of-codecraft: This is me not having automatic go formatting on this project.  I use it on others so I get lazy as normally when I save the file it just fixes it all.  I will take care of it.

libs/digger_config/digger_config.go
ben-of-codecraft: Yep, you are right will adjust.

libs/digger_config/digger_config.go
```diff
@@ -269,15 +269,22 @@ func HandleYamlProjectGeneration(config *DiggerConfigYaml, terraformDir string,
...
+						tgParsingConfig := TerragruntParsingConfig{
 							CreateProjectName: true,
 							DefaultWorkflow:   workflow,
 							WorkflowFile:      b.WorkflowFile,
 							FilterPath:        path.Join(terraformDir, *b.RootDir),
-						}, terraformDir)
+						};						
+
+						// allow blocks to pass in roles that can be assummed by aws 					
+						tgParsingConfig.AwsRoleToAssume = b.AwsRoleToAssume
```
motatoes, line 280: sorry another nit, why not just pass it in the struct above directly? :D

motatoes (approved):

PR #1721, opened 2024-09-23
Feature: Added in skip_merge_check option to workflow configuration

ben-of-codecraft:
This adds a new configuration option to workflow configuration, `skip_merge_check` this override default behavior of Digger to check if a PR is in a mergeable state before allowing commands like `terraform apply`. 

Our use case is we keep multiple environments in the same repo and some environments and certain accounts should have branch protection controls in place, which exist at a repo level not at a directory level.  This option allows us to let developer environments bypass protections or other rules and apply state, even if the branch has not passed checks. 

Usage:
```yaml
    workflow_configuration:
      on_pull_request_pushed: ["digger plan"]
      on_pull_request_closed: ["digger unlock"]
      on_commit_to_default: ["digger unlock"]
      skip_merge_check: true                                    # skips the pr mergeability check before running apply command
```

### DIFF
--- a/libs/ci/github/github.go
+++ b/libs/ci/github/github.go
@@ -500,6 +500,7 @@ func ConvertGithubPullRequestEventToJobs(payload *github.PullRequestEvent, impac
 				RequestedBy:        *payload.Sender.Login,
 				CommandEnvProvider: CommandEnvProvider,
 				StateEnvProvider:   StateEnvProvider,
+				SkipMergeCheck: 	skipMerge,
 			})
 		} else if *payload.Action == "closed" {
 			jobs = append(jobs, scheduler.Job{
@@ -521,6 +522,7 @@ func ConvertGithubPullRequestEventToJobs(payload *github.PullRequestEvent, impac
 				RequestedBy:        *payload.Sender.Login,
 				CommandEnvProvider: CommandEnvProvider,
 				StateEnvProvider:   StateEnvProvider,
+				SkipMergeCheck: 	skipMerge,
 			})
 		} else if *payload.Action == "converted_to_draft" {
 			var commands []string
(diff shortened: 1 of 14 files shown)


libs/ci/github/github.go
```diff
@@ -471,6 +478,7 @@ func ConvertGithubPullRequestEventToJobs(payload *github.PullRequestEvent, impac
 				RequestedBy:        *payload.Sender.Login,
 				CommandEnvProvider: CommandEnvProvider,
 				StateEnvProvider:   StateEnvProvider,
+				SkipMergeCheck: 	skipMerge,
```
motatoes, line 481: I think you should also added it in the rest of the branch tree, any line where it says jobs = append() should also have the line for SkipMergeCheck field

libs/ci/github/github.go
```diff
@@ -445,6 +445,13 @@ func ConvertGithubPullRequestEventToJobs(payload *github.PullRequestEvent, impac
 			return nil, false, fmt.Errorf("failed to find workflow config '%s' for project '%s'", project.Workflow, project.Name)
 		}
 
+		var skipMerge bool
```
motatoes, line 448: we are also missing the same logic in `ConvertIssueCommentEventToJobs` function so that it is always passed to the job struct correctly

motatoes (changes requested):

libs/ci/github/github.go
ben-of-codecraft: Added the logic in here

libs/ci/github/github.go
ben-of-codecraft: Wen through and added the logic plus the option to all locations where a Job struct is created.

motatoes (approved):

Issue #1109, opened 2024-02-01
Print the version + commit ID at the top of every binary invocation


motatoes:
It will be useful to print out the git commit ID of every build, for example:

```
Running digger v0.3.22 (commit: #5ea3fa)
```

PR #1770, opened 2024-10-22
Now shows commit id + version

SSROCKS30:
This PR fixes issue #1109 Print commit id and version at the top of every binary invocation

## Changes
- Modified `cli/cmd/digger/main.go` to include version and commit ID printing logic
- Updated `cli/cmd/digger/main_test.go` to account for the new output in tests
- Updated `backend/Makefile` to include version and commit ID during the build process

## Implementation Details
- Added version and commit ID printing at the start of the `main()` function in `main.go`
- The version information is retrieved from build-time variables set in the Makefile
- Dockerfiles have been updated to pass the necessary build arguments

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -4,3 +4,7 @@ export
 
 start:
 	go run main.go
+
+# Add or modify a build target
+build:
+	go build -ldflags "-X main.version=$(shell git describe --tags) -X main.commitID=#$(shell git rev-parse --short HEAD)" -o digger cli/cmd/digger/main.go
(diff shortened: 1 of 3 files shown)


motatoes:
Hi there thanks for contributing, unfortunately the idea of this PR is not to introduce a new command rather we want to print this in the default entrypoint of the program so directly in the main value. I also am not sure if we correctly override the value of the "version" variable during the build so that needs to be verified as well

motatoes (changes requested):
Please make it print these values in the main function not as a seperate command

PR #1803, opened 2024-11-08
WIP add github connections to EE

motatoes:
support for multiple orgs from a single backend


backend/utils/crypt.go
```diff
@@ -0,0 +1,140 @@
...
+	"crypto/cipher"
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+	"github.com/diggerhq/digger/backend/models"
+	"io"
+)
+
+// Encrypt encrypts a plaintext string using AES-256-GCM
+func AESEncrypt(key []byte, plaintext string) (string, error) {
```
ZIJ, line 14: its odd that we need own functions for encrypting / decripting strings with AES - isn't there a library?

ZIJ (approved):

backend/utils/crypt.go
motatoes: yes we are using the native crypto library in golang but as everything else in go its verbose enough to warrant a wrapper function, encryption is multiple steps and there is no AESEncrypt() that would just do it, not that I'm aware of :)

PR #1843, opened 2024-11-25
Fix drift check logic for no chnages

chiragsoni-employ:
## Description
Terraform show command output json can contain `output_changes` even if there are no changes required in plan, it will list all those changes with action `no-op`, so a normal `nil` check will not work with this, Due to this issue current code is resulting into false positives where drift checker will send notifications even if there are no changes detected

## Solution
Here I have checked for action in `output_changes` in a similar way we do in `resource_changes`


motatoes (approved):

motatoes:
Hi there many thanks for this contribution! @chiragsoni-employ

PR #1871, opened 2025-01-15
fix: auto_merge not working properly with orchestrator backend

duy0611:
When running with Digger backend mode, PR is merged when `auto_merge` is enabled even though not all impacted projects have been applied. This is caused because Digger batch does not have context whether all the needed jobs have been triggered when using `digger apply` with `-p` flag. 

Unlike in backendless mode, where all the jobs are run sequentially and the parent one does have context if all the impacted projects have been applied, in backend mode jobs are triggered via the Batch one, which does not have this context. This PR addresses the issue by adding one more column to store this context, which will then be used in the subsequent run.

## Summary by CodeRabbit

- **New Features**
  - Added a new setting to track whether all impacted projects are covered in batch operations, visible in batch-related actions and behaviors.

- **Bug Fixes**
  - Improved handling of batch auto-merge conditions to ensure merges only occur when all impacted projects are included.

- **Tests**
  - Enhanced test coverage for batch creation and auto-merge scenarios, including new conditions for impacted project coverage.

- **Chores**
  - Updated database schema to support the new impacted projects tracking field.
  - Refined internal logic to consistently handle the new batch coverage setting across the platform.


motatoes:
Hi @duy0611 thanks for much for this PR! Please allow me a few days to review it as im currently loaded with stuff

motatoes:
hi @duy0611 sorry I just got to start reviewing now! I took a quick look and I have a few small comments:

- Firstly we need to ensure that the atlas migrations file is in the right order since I think other migrations have been merged since, we would need to delete and regenerate migrations after fixing merge conflicts
- We need to fix merge conflicts with develop branch unfrotunately!
- I think some changes from the libs/ folder are missing in this PR, I saw some changes to the signature of `ConvertGithubPullRequestEventToJobs` function but I don't see the respective changes in the function, could you please review if some changes have not been commited?

motatoes:
Hi @duy0611 ! unfortunately I came back too late to this PR and now there are merge conflicts. I have reviewed the changes in general and I have some comments:

- I think some changes to libs/ are not present in this contribution. Could you please that libs/ changes is also in there? Specifically the changes you made to `ConvertGithubPullRequestEventToJobs`
- Regarding the migration, I think it needs to be rebased from develop and re-generated to preserve migration fils order!

Looking forward to your updated contribution!

duy0611:
Hi @motatoes . I updated my PR now to reflect your comments. Could you review that again? Thank you.

motatoes (approved):

PR #1930, opened 2025-03-31
docs: describe a safer approach to using pinned version

birjj:
As exemplified by the recent [`tj-actions/changed-files` compromise](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised#summary-of-the-incident), pinning GitHub Action references to a tag is inherently unsafe: if the referenced action is compromised, the attacker can update any tags, including old ones, to point to whatever they want.

For something as crucial as IaC management it'd probably be preferable to use _actually_ pinned versions over pinning to mutable tags. This PR updates the [_"Specifying version"_](https://docs.digger.dev/ce/howto/versioning#use-a-pinned-version) docs to describe that process instead. Might as well push people towards the safer workflow 🤷

Also includes a fix to the docs so the code example for `vLatest` actually uses `@vLatest`, and removes the `description` MDX header, since it contains text that's immediately replicated in a more-visible warning directly below it.

Feel free to make any changes you feel relevant to make the wording flow better! This PR is mostly to suggest that the pin-to-commit-SHA approach is described in the docs.

## Summary by CodeRabbit

- **Documentation**
	- Updated versioning guidance for the Digger action.
	- Revised instructions now recommend using a commit SHA for pinning releases, enhancing security and performance.
	- Clarified the usage of the "vLatest" tag to denote the most recent release with pre-built binaries.

### DIFF
--- a/docs/ce/howto/versioning.mdx
+++ b/docs/ce/howto/versioning.mdx
@@ -4,9 +4,7 @@ description: "For serious usecases always use a pinned version which is of the f
 ---
 
 <Warning>
-  For serious usecases always use a pinned version which is of the form @vX.Y.Z
-  since this will download a compiled binary. In addition to being faster to run, it
-  is also more secure than using a commit from a branch
+  For serious usecases always use a commit SHA reference for the digger action, of the format `@<40-character-commit-SHA>`. We recommend using a commit SHA from one of [our releases](https://github.com/diggerhq/digger/releases), since this will download a compiled binary. In addition to being faster to run, it is also more secure than using a commit from a branch or tag.
 </Warning>
 
 ## Use vLatest tag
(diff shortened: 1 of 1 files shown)


motatoes:
Hi @birjj thanks for this! I totally agree with hardcoding the commit. The issue is that our current check for whether to build or download binary is based on if it [starts with the letter "v"](https://github.com/diggerhq/digger/blob/fdb01cc8bcc395220e3f8976968bba388aff23cb/action.yml#L419) to identify release. So we need to figure out a better way of saying download binary vs build from source in the composite action before this can work with pure commit sha ref.

I think We need to figure that out before updating docs, (unless I missed something)

birjj:
Ah, yes that would definitely be a problem. I don't have any good ideas for getting a reference to the release from the commit SHA, so that's a blocker.

motatoes:
Thanks, I've created a backlog issue to look into that so we can address it and then get back to this in the near future :D

PR #2082, opened 2025-08-14
remove redundant logs

motatoes:
This logline is clogging the backend debug logs and offers no value, removing


breardon2011 (approved):
Looks good.

PR #2139, opened 2025-09-17
move taco back in, minus docs

breardon2011:


motatoes (commented):
remove the taco/taco binary

motatoes (approved):

Issue #2108, opened 2025-08-22
check for nil pointer exception and nil error in function call to GetGithubAppInstallationLink


motatoes:
for later, we saw this type of NPE exception 

```
5-08-22 20:11:07.040 BST
Stack trace:
2025-08-22 20:11:07.040 BST
goroutine 51 [running]:
2025-08-22 20:11:07.040 BST
runtime/debug.Stack()
2025-08-22 20:11:07.040 BST
/usr/local/go/src/runtime/debug/stack.go:26 +0x5e
2025-08-22 20:11:07.040 BST
github.com/diggerhq/digger/backend/controllers.handlePullRequestEvent.func1()
2025-08-22 20:11:07.040 BST
/go/src/github.com/diggerhq/digger/backend/controllers/github.go:458 +0x3f
2025-08-22 20:11:07.040 BST
panic({0x2705480?, 0x4acee50?})
2025-08-22 20:11:07.040 BST
/usr/local/go/src/runtime/panic.go:787 +0x132
2025-08-22 20:11:07.040 BST
github.com/diggerhq/digger/backend/controllers.handlePullRequestEvent({0x32f01a8, 0x4d44560}, 0xc00050e100, {0x32bf5a0, 0x4d44560}, 0x5fc7c)
2025-08-22 20:11:07.040 BST
/go/src/github.com/diggerhq/digger/backend/controllers/github.go:507 +0x8f3
2025-08-22 20:11:07.040 BST
github.com/diggerhq/digger/backend/controllers.DiggerController.GithubAppWebHook.func3({0x32f52f8?, 0xc000a59bf0?})
2025-08-22 20:11:07.040 BST
/go/src/github.com/diggerhq/digger/backend/controllers/github.go:146 +0x89
2025-08-22 20:11:07.040 BST
created by github.com/diggerhq/digger/backend/controllers.DiggerController.GithubAppWebHook in goroutine 16
2025-08-22 20:11:07.040 BST
/go/src/github.com/diggerhq/digger/backend/controllers/github.go:144 +0xffc
2025-08-22 20:11:07.040 BST
```

Related to `link` being nil here:

```

	link, err := models.DB.GetGithubAppInstallationLink(installationId)
	if err != nil {
		slog.Error("Error getting GitHub app installation link",
			"installationId", installationId,
			"error", err,
		)
		return fmt.Errorf("error getting github app link")
	}
	organisationId := link.OrganisationId

```

[...]

PR #2191, opened 2025-09-20
Fix nil pointer dereference in GetGithubAppInstallationLink calls

MrKeiKun:
**Description:**
The GetGithubAppInstallationLink function in digger/backend/models/storage.go returns nil, nil when no record is found, which is a valid scenario when a GitHub app installation hasn't been linked to an organization yet.

**Reference:** #2108


motatoes:
Thanks @MrKeiKun !!

PR #2258, opened 2025-09-24
workspace tags wip

breardon2011:


breardon2011:
closing because I think the DB driver document query POC's will impact this implementation if successful. given the clunky trino one was still faster than our current I am prone to think at least one of them will be.

PR #2265, opened 2025-09-29
DB Query Engine 

breardon2011:
Adds modular DB backing, allow bucketstore abstraction. 

Note: a large amount of the LOC change is the test suite and data.

### DIFF
--- a/taco/internal/domain/interfaces.go
+++ b/taco/internal/domain/interfaces.go
@@ -55,10 +55,9 @@ type UnitManagement interface {
 // ============================================
 
 // TFEOperations defines what TFE handler needs.
-// TFE needs to auto-create workspaces but not list/delete/version operations.
+// TFE needs read/write/lock operations only
 type TFEOperations interface {
 	StateOperations
-	Create(ctx context.Context, id string) (*storage.UnitMetadata, error)
 }
 
 // ============================================
--- a/taco/internal/api/routes.go
+++ b/taco/internal/api/routes.go
@@ -37,7 +37,7 @@ type Dependencies struct {
 // RegisterRoutes registers all API routes with interface-scoped dependencies.
 // - Backend: StateOperations (6 methods) - cannot create/list/delete/version
 // - S3-compat: StateOperations (6 methods) - cannot create/list/delete/version
-// - TFE: TFEOperations (7 methods) - can create but not list/delete/version
+// - TFE: TFEOperations (6 methods) - read/write/lock only
 // - Unit: UnitManagement (11 methods) - full access for management API
 // - RBAC: RBACManager + QueryStore - NO unit access at all
 func RegisterRoutes(e *echo.Echo, deps Dependencies) {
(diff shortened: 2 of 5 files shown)


taco/cmd/statesman/main.go
```diff
@@ -12,50 +12,150 @@ import (
...
 	default:
-		store = storage.NewMemStore()
+		blobStore = storage.NewMemStore()
 		log.Printf("Using in-memory storage")
 	}
 
-	// Initialize analytics with system ID management (always create system ID)
-	analytics.InitGlobalWithSystemID("production", store)
 
+	// 3. Create the base OrchestratingStore
```
motatoes: 3. is redundant in comment, where is 1 and 2?

taco/cmd/statesman/main.go
```diff
@@ -12,50 +12,150 @@ import (
...
+	orchestratingStore := storage.NewOrchestratingStore(blobStore, queryStore)
```
motatoes: this orchestrating store is grouping blobStore and queryStore which are both unrelated it seems, just use UnitStore and QueryStore independently ?

taco/internal/domain/interfaces.go
```diff
@@ -0,0 +1,147 @@
...
+	ListVersions(ctx context.Context, id string) ([]*storage.VersionInfo, error)
+	RestoreVersion(ctx context.Context, id string, versionTimestamp time.Time, lockID string) error
+}
+
+// ============================================
+// TFE-Specific Operations
+// ============================================
+
+// TFEOperations defines what TFE handler needs.
+// TFE needs to auto-create workspaces but not list/delete/version operations.
```
motatoes: is this true that we "auto-create" I thought that we dont auto-create but rather we require them to be created before any operation on them would succeed

taco/internal/query/sqlite/TEST_GUIDE.md
```diff
@@ -0,0 +1,601 @@
+# Test Suite Guide - Quick Reference
```
motatoes, line 1: is this doc adding any value? maybe an explanation on running the tests but what's summarising the number of tests will immediately go out of sync with the code. I would remove all but example commands on how to run the tests

taco/internal/observability/sync_health.go
```diff
@@ -0,0 +1,67 @@
+package observability
```
motatoes, line 1: I didn't quite understand this and the other sync methods in query/internal/sql_store.go

motatoes (approved):

taco/internal/query/sqlite/TEST_GUIDE.md
breardon2011: yeah thats a good point, sort of made it for myself initially but through that lens it should probably be minimal i agree

taco/internal/domain/interfaces.go
breardon2011: yeah thats a good catch i'll confirm the behaviour now and adjust

breardon2011: we do prevent auto create in most but there was one where not it turns out (createstateversiondirect), I'll adjust both that method and the comment

PR #2277, opened 2025-10-01
chore(run-digger-action): pin all action dependencies to commit hash (no-op)

sidpalas:
This PR pins all actions consumed within the run-digger action using commit hashes.

I honored the existing versions (using the latest allowable version as specified in the current action) so this should be a no-op.

Why?
> The only way to ensure the same code is running for GitHub Actions with each run is to pin to the commit hash. 

> This change will decrease the likelihood of an unexpected break due to a sub-action changing, but more importantly, helps defend against potential supply chain attacks such as https://www.wiz.io/blog/github-action-tj-actions-changed-files-supply-chain-attack-cve-2025-30066

---

TODO: merge/rebase on #2278 to separate the formatting changes from the dependency pinning changes

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -7,23 +7,23 @@ inputs:
   local-dev-mode:
     description: run digger for local development?
     required: false
-    default: 'false'
+    default: "false"
   local-dev-cli-path:
     description: The path to where the compiled digger cli on the self-hosted runner exists (absolute path)
     required: false
-    default: './digger'
+    default: "./digger"
   ee:
     description: use ee cli?
     required: false
-    default: 'false'
+    default: "false"
   fips:
     description: build with fips140 standard?
     required: false
-    default: 'false'
+    default: "false"
   setup-aws:
     description: Setup AWS
     required: false
-    default: 'false'
+    default: "false"
   aws-access-key-id:
     description: AWS access key id
     required: false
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -7,7 +7,7 @@ inputs:
   local-dev-mode:
     description: run digger for local development?
     required: false
-    default: false
+    default: 'false'
```
sidpalas: This was a drive-by fix. Composite actions can only have string inputs.

action.yml
```diff
@@ -204,33 +204,32 @@ inputs:
     required: false
     default: ''
   cache-dependencies:
-    description: "Leverage actions/cache to cache dependencies to speed up execution"
+    description: 'Leverage actions/cache to cache dependencies to speed up execution'
     required: false
     default: 'false'
```
sidpalas: There is a mix of single and double quotes for strings... do you want me to leave as is or standardize?

action.yml
```diff
@@ -464,28 +462,28 @@ runs:
...
-          if [[ ${{ inputs.fips }} == "true" ]]; then
-            export GODEBUG=fips140=only
-            export GOFIPS140=v1.0.0
-          fi        
-          go build -o digger ./cmd/digger
-          chmod +x digger
-          PATH=$PATH:$(pwd)
-          cd $GITHUB_WORKSPACE
-          digger
+        if [[ ${{ inputs.ee }} == "true" ]]; then
```
sidpalas, line 473: Not sure why this script was indented more than the others...

Maybe I should create a separate PR with all these auto-formatting fixes?

PR #2283, opened 2025-10-02
feat(run-digger-cli): add digger-version input

sidpalas:
Adds support for an optional `digger-version` input which enables pinning usage of the digger action to the commit hash without incurring the performance penalty of needing to rebuild from source.

I don't love how complex the conditional logic is here. Trying to think how to potentially simplify this logic.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -438,7 +438,7 @@ runs:
       shell: bash
 
     - name: build and run digger
-      if: ${{ inputs.digger-version == "" && !startsWith(github.action_ref, 'v') && inputs.local-dev-mode == 'false' }}
+      if: ${{ inputs.digger-version == '' && !startsWith(github.action_ref, 'v') && inputs.local-dev-mode == 'false' }}
       shell: bash
       env:
         PLAN_UPLOAD_DESTINATION: ${{ inputs.upload-plan-destination }}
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -257,7 +261,7 @@ runs:
         exit 1
       shell: bash
       if: inputs.setup-google-cloud == 'true'
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
```
sidpalas, line 288: This is a drive by fix for one action I missed in #2277

action.yml
```diff
@@ -574,7 +578,7 @@ runs:
     - name: run digger in local dev mode
       if: ${{ inputs.local-dev-mode == 'true' }}
       env:
-        actionref: ${{ github.action_ref }}
+        DIGGER_VERSION: ${{ github.action_ref }}
```
sidpalas, line 605: Updated here for consistency, but I don't actually think this gets used...

sidpalas:
@motatoes -- anything else you think needs to be done here before it is ready to merge?

motatoes (approved):

motatoes:
Thanks @sidpalas ! merging 🥳

PR #2305, opened 2025-10-07
fallback strategy, remove sleep block

breardon2011:
Handles condition where post merge action fails because cannot clone non existent branch, now uses fallback if cannot grab target branch of PR. 
Screenshot 2025-10-07 at 2 54 22 PM

### DIFF
--- a/backend/controllers/github_helpers.go
+++ b/backend/controllers/github_helpers.go
@@ -694,7 +694,7 @@ func getDiggerConfigForPR(gh utils.GithubClientProvider, orgId uint, prLabels []
 	}
 
 	var prBranch string
-	var targetBranch string // Use PR's actual target branch instead of repo default
+	var targetBranch string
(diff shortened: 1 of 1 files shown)


libs/ci/github/github.go
```diff
@@ -498,6 +498,14 @@ func (svc GithubService) CheckBranchExists(branchName string) (bool, error) {
...
 }
 
+func (svc GithubService) GetDefaultBranch() (string, error) {
+	repo, _, err := svc.Client.Repositories.Get(context.Background(), svc.Owner, svc.RepoName)
+	if err != nil {
+		return "", fmt.Errorf("error getting repository: %v", err)
+	}
+	return repo.GetDefaultBranch(), nil
+}
+
```
motatoes: unused function ?

breardon2011: Removed

PR #2312, opened 2025-10-09
Change "fallback" to skip under fallback condition 

breardon2011:
When a `branch not found after merge` is detected, an empty graph is returned to skip the runs.

Fallback logic / references also removed

### DIFF
--- a/backend/controllers/github_comment.go
+++ b/backend/controllers/github_comment.go
@@ -2,7 +2,6 @@ package controllers
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"log/slog"
 	"os"


backend/controllers/github_helpers.go
```diff
@@ -591,79 +591,58 @@ func retrieveConfigFromCache(orgId uint, repoFullName string) (string, *digger_c
...
 			}
-
-			slog.Info("Successfully loaded config from target branch",
-				"targetBranch", targetBranch,
-				"repoFullName", repoFullName,
-			)
-
-			actualBranchUsed = targetBranch
+			emptyGraph := graph.New(projectHash, graph.Directed())
+			return "", ghService, emptyConfig, emptyGraph, nil
```
motatoes: shouldn't we be returning a special error here like ErrorBranchNotFound to be able to catch it in the caller?

backend/controllers/github_helpers.go
```diff
@@ -591,79 +591,58 @@ func retrieveConfigFromCache(orgId uint, repoFullName string) (string, *digger_c
 	return repoCache.DiggerYmlStr, &config, &projectsGraph, &taConfig, nil
 }
 
-// GetDiggerConfigForBranchWithFallback attempts to load config from the specified branch,
-// and falls back to the target branch if the branch doesn't exist and shouldFallback is true
-func GetDiggerConfigForBranchWithFallback(gh utils.GithubClientProvider, installationId int64, repoFullName string, repoOwner string, repoName string, cloneUrl string, branch string, targetBranch string, shouldFallback bool, changedFiles []string, taConfig *tac.AtlantisConfig) (string, *github2.GithubService, *digger_config.DiggerConfig, graph.Graph[string, digger_config.Project], error) {
```
motatoes, line 596: we should probably change the name of this function to say GetDiggerConfigForBranchWithGracefulHandling or similar

backend/controllers/github_helpers.go
```diff
@@ -773,39 +742,35 @@ func getDiggerConfigForPR(gh utils.GithubClientProvider, orgId uint, prLabels []
...
 		)
 	}
 
-	// Use the fallback method with merge-based fallback logic
+	// get config 
 	diggerYmlStr, ghService, config, dependencyGraph, err := GetDiggerConfigForBranchWithFallback(
 		gh, installationId, repoFullName, repoOwner, repoName, cloneUrl, 
-		prBranch, targetBranch, shouldFallback, changedFiles, taConfig,
+		prBranch, isMerged, changedFiles, taConfig,
 	)
```
motatoes, line 778: I believe if failure to clone happens here we can return right away, no need to process all these empty structs unnecessarly

backend/controllers/github_comment.go
```diff
@@ -163,6 +164,14 @@ func handleIssueCommentEvent(gh utils.GithubClientProvider, payload *github.Issu
 
 	diggerYmlStr, ghService, config, projectsGraph, prSourceBranch, commitSha, changedFiles, err := getDiggerConfigForPR(gh, orgId, prLabelsStr, installationId, repoFullName, repoOwner, repoName, cloneURL, issueNumber)
 	if err != nil {
+		if errors.Is(err, ErrBranchNotFoundPostMerge) {
```
motatoes: Is this needed? That's not the graceful handling function being called here in the comment function

breardon2011: oh yeah wrong caller

PR #2319, opened 2025-10-15
Webhook auth and internal endpoints

breardon2011:
Add organizations 
Add internal endpoints, org scoped with webhook auth

### DIFF
--- a/taco/internal/api/org_handler.go
+++ b/taco/internal/api/org_handler.go
@@ -14,7 +14,7 @@ import (
 
 // OrgHandler handles organization and user management endpoints
 type OrgHandler struct {
-	orgRepo     domain.OrganizationRepository // ✅ Interface, not concrete implementation
+	orgRepo     domain.OrganizationRepository 
 	userRepo    domain.UserRepository
 	rbacManager *rbac.RBACManager
 }


taco/internal/api/internal.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/diggerhq/digger/opentaco/internal/middleware"
+	"github.com/diggerhq/digger/opentaco/internal/rbac"
+	"github.com/diggerhq/digger/opentaco/internal/repositories"
+	unithandlers "github.com/diggerhq/digger/opentaco/internal/unit"
+	"github.com/labstack/echo/v4"
+)
+
+
+func RegisterInternalRoutes(e *echo.Echo, deps Dependencies) {
+	webhookSecret := os.Getenv("OPENTACO_WEBHOOK_SECRET")
```
motatoes, line 18: I think we should make it more explicit like a flag called OPENTACO_ENABLE_INTERAL_ENDPOINTS since the flag OPENTACO_WEBHOOK_SECRET doesn't really tell me that it will lead to alot of admin endpoints are going to be exposed

taco/internal/api/internal.go
```diff
@@ -0,0 +1,210 @@
...
+	var orgRepo domain.OrganizationRepository
+	var userRepo domain.UserRepository
+	
+	if deps.QueryStore != nil {
+		orgRepo = repositories.NewOrgRepositoryFromQueryStore(deps.QueryStore)
+		userRepo = repositories.NewUserRepositoryFromQueryStore(deps.QueryStore)
+	}
+
+	// Create internal group with webhook auth (with orgRepo for existence check)
+	internal := e.Group("/internal")
```
motatoes, line 36: maybe /internal/api is clearer as a prefix route

taco/internal/api/routes.go
```diff
@@ -274,4 +274,7 @@ func RegisterRoutes(e *echo.Echo, deps Dependencies) {
 			}},
 		})
 	})
+
+	// Register webhook-authenticated internal routes (if OPENTACO_WEBHOOK_SECRET is set)
```
motatoes, line 278: if you end up changing the flag to OPENTACO_ENABLE_INTERAL_ENDPOINTS then this comment needs updating

motatoes (approved):

PR #2327, opened 2025-10-18
feat/ui units

motatoes:
- **add units elements**
- **implement mocked units**
- **fix units**
- **add workos webhook and id**

### DIFF
--- a/taco/internal/domain/organization.go
+++ b/taco/internal/domain/organization.go
@@ -14,7 +14,7 @@ var (
 	ErrInvalidOrgID = errors.New("invalid organization ID format")
 )
 
-// OrgIDPattern defines valid organization ID format: lowercase alphanumeric, hyphens, underscores
+// OrgIDPattern defines valid organization ID format: alphanumeric, hyphens, underscores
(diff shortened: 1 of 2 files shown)


taco/internal/domain/organization.go
```diff
@@ -15,7 +15,7 @@ var (
 )
 
 // OrgIDPattern defines valid organization ID format: lowercase alphanumeric, hyphens, underscores
```
breardon2011: should update comment as well

ui/src/routes/_authenticated/_dashboard/dashboard/units.index.tsx
```diff
@@ -0,0 +1,243 @@
...
+)({
+  component: RouteComponent,
+  loader: async ({ context }) => {
+    const { user, organisationId } = context;
+    const unitsData = await listUnitsFn({data: {organisationId: organisationId || '', userId: user?.id || '', email: user?.email || ''}})
+    return { unitsData: unitsData, user, organisationId }
+  }
+})
+
+// Mock data - replace with actual data fetching
```
breardon2011: looks like these are not used now should probably remove if so

breardon2011 (approved):
LGTM, a couple comments but nothing blocking

PR #2331, opened 2025-10-22
fix/fix build and routes

motatoes:
- **fix build**
- **add more routes**
- **implement routes**

For github app endpoints we add "orchestrator/" at the start:

github callback: "/orchestrator/github/callback"
github webhook: "/orchestrator/github/webhook"

### DIFF
--- a/ui/src/components/DiggerYmlButton.tsx
+++ b/ui/src/components/DiggerYmlButton.tsx
@@ -2,7 +2,6 @@
 
 import { Button } from "@/components/ui/button"
 import { FileText } from "lucide-react"
-import { trackDiggerYmlAdded } from "@/lib/analytics"
 
 interface DiggerYmlButtonProps {
   onClick: () => void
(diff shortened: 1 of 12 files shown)


ui/src/routes/api/auth/workos/switch-org.tsx
```diff
@@ -30,7 +30,7 @@ async function refreshSession({
   const session = await getSessionFromCookie();
   if (!session) {
     if (ensureSignedIn) {
-      await redirectToSignIn();
+      // await redirectToSignIn();
```
breardon2011, line 33: Commented out code

breardon2011 (approved):
LGTM

ui/src/routes/api/auth/workos/switch-org.tsx
motatoes: yeah this part doesn't work yet but it will be implemented in the next PR

PR #2332, opened 2025-10-22
fix: fix terragrunt non-interactive flag

sidpalas:
Terragrunt redesigned their CLI (released in [v0.77.22](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.77.22), see [GH issue comment](https://github.com/gruntwork-io/terragrunt/issues/3445#issuecomment-2819076926)) to remove the `terragrunt` prefix.

This PR removes the `--terragrunt-non-interactive` for `apply` and `destroy` commands. This flag is not valid for newer versions of Terragrunt and was duplicative with ENV vars set here: https://github.com/diggerhq/digger/blob/53173e54b1e4e1638faad6f421e808c05490b2d5/libs/execution/terragrunt.go#L123-L125

By removing the CLI flag but setting both `TERRAGRUNT_NON_INTERACTIVE=true` AND `TG_NON_INTERACTIVE=true`, digger remains compatible with versions before and after the CLI redesign https://terragrunt.gruntwork.io/docs/migrate/cli-redesign/

---

> [!IMPORTANT]  

---

- [x] I did **not** use AI tools in this contribution.
- [ ] I used AI tools and have disclosed details below.

**Details (if applicable):**
> _Example: Used ChatGPT to help with doc phrasing._  
> _Example: Code generated by Copilot; reviewed and verified manually._

---

#### 💡 Notes

### DIFF
--- a/libs/execution/terragrunt.go
+++ b/libs/execution/terragrunt.go
@@ -26,7 +26,6 @@ func (terragrunt Terragrunt) Init(params []string, envs map[string]string) (stri
 func (terragrunt Terragrunt) Apply(params []string, plan *string, envs map[string]string) (string, string, error) {
 	params = append(params, []string{"-lock-timeout=3m"}...)
 	params = append(params, "--auto-approve")
-	params = append(params, "--non-interactive")
 	if plan != nil {
 		params = append(params, *plan)
 	}
@@ -40,7 +39,6 @@ func (terragrunt Terragrunt) Apply(params []string, plan *string, envs map[strin
 
 func (terragrunt Terragrunt) Destroy(params []string, envs map[string]string) (string, string, error) {
 	params = append(params, "--auto-approve")
-	params = append(params, "--non-interactive")
 	stdout, stderr, exitCode, err := terragrunt.runTerragruntCommand("destroy", true, envs, nil, params...)
 	if exitCode != 0 {
 		logCommandFail(exitCode, err)
--- a/action.yml
+++ b/action.yml
@@ -91,7 +91,7 @@ inputs:
   terragrunt-version:
     description: Terragrunt version
     required: false
-    default: v0.90.0
+    default: v0.73.7
   opentofu-version:
     description: OpenTofu version
     required: false


ZIJ:
Actually @sidpalas I just realised that this would be a breaking change for people who set terragrunt version explicitly to anything older than v0.77.2

Spoke with @motatoes and he suggested setting equivalent env vars (`TG_NON_INTERACTIVE` and older equivalent `TERRAGRUNT_NON_INTERACTIVE`) - this way we would not have to do a conditional check of the actual terragrunt version

sidpalas:
> Actually @sidpalas I just realised that this would be a breaking change for people who set terragrunt version explicitly to anything older than v0.77.2
> 
> 
> 
> Spoke with @motatoes and he suggested setting equivalent env vars (`TG_NON_INTERACTIVE` and older equivalent `TERRAGRUNT_NON_INTERACTIVE`) - this way we would not have to do a conditional check of the actual terragrunt version
> 
> 

Yeah that makes sense as a way to avoid the breaking change.

It would be nice to land https://github.com/diggerhq/digger/pull/2287 before making that change to avoid triplicating the env vars...

Alternatively, 
- the env var could be set that the job level (even though it is only needed at the step where digger is executed)
- the env var could be set within the go code

---

Actually that third option may be the best since it doesn't need to be configurable and doesn't need to be exposed to the end users of the action.

motatoes:
> Actually that third option may be the best since it doesn't need to be configurable and doesn't need to be exposed to the end users of the action.

@sidpalas yes that would be my preference indeed, setting it at the go code level, there is an envs map in the same function where we can set it so it is exposed directly to the terragrunt process, should be a 2-liner change!

sidpalas:
> > Actually that third option may be the best since it doesn't need to be configurable and doesn't need to be exposed to the end users of the action.
> 
> 
> 
> @sidpalas yes that would be my preference indeed, setting it at the go code level, there is an envs map in the same function where we can set it so it is exposed directly to the terragrunt process, should be a 2-liner change!

Looks like they are already set!

https://github.com/sidpalas/digger/blob/sp/fix-terragrunt-flags/libs/execution/terragrunt.go#L122-L125

So we just need to remove the hardcoded cli flag

motatoes:
@sidpalas oh nice, yeah its a 1 liner then haha

libs/execution/terragrunt.go
```diff
@@ -26,7 +26,6 @@ func (terragrunt Terragrunt) Init(params []string, envs map[string]string) (stri
 func (terragrunt Terragrunt) Apply(params []string, plan *string, envs map[string]string) (string, string, error) {
 	params = append(params, []string{"-lock-timeout=3m"}...)
 	params = append(params, "--auto-approve")
-	params = append(params, "--terragrunt-non-interactive")
```
sidpalas, line 29: This is duplicative with ENV var set here: https://github.com/diggerhq/digger/blob/53173e54b1e4e1638faad6f421e808c05490b2d5/libs/execution/terragrunt.go#L123-L125

By removing the CLI flag but setting both `TERRAGRUNT_NON_INTERACTIVE=true` AND `TG_NON_INTERACTIVE=true` this remains compatible with versions before and after the CLI redesign https://terragrunt.gruntwork.io/docs/migrate/cli-redesign/

sidpalas:
Updated to remove the CLI flag in favor of the environment variables (which were already being set)

PR #2391, opened 2025-11-05
fix posthog fetching

motatoes:
used cursor in implementation of this pr but all validated - the issue with VITE_PUBLIC is that it needs to exist at build time which is annoying - we instead rely on server config pattern to be able to pickup posthog env vars at runtime


breardon2011 (commented):
LGTM

PR #2420, opened 2025-11-07
fix repo name

motatoes:
"/" is not allowed in the name and rejected by k8s manifest

used gpt to generate the sanitization function and its tests


breardon2011 (approved):
approve

PR #2451, opened 2025-11-14
Remote runs POC (backend) 

breardon2011:
defaults units to remote for testing, the UI could explicity set local or remote accordingly but this way i could create a unit in the ui without changing the ui

Implemented remote runs that run on the server using temp dir 

User's Terraform → cloud block → points to workspace 22fb011f-...
Backend downloads → state for orgID/unitID (same workspace!)
Backend modifies → removes cloud block in temp dir
Backend runs terraform → uses local state file (same content!)
Backend uploads → state back to orgID/unitID (same location!)
User's next run → cloud block → fetches updated state 

> [!IMPORTANT]  

---

- [ ] I did **not** use AI tools in this contribution.
- [x] I used AI tools and have disclosed details below.

**Details (if applicable):**
Cursor in a directed way "do x, do y" 

---

#### 💡 Notes

### DIFF
--- a/ui/src/components/UnitCreateForm.tsx
+++ b/ui/src/components/UnitCreateForm.tsx
@@ -32,14 +32,9 @@ export default function UnitCreateForm({
   const [unitType, setUnitType] = React.useState<'local' | 'remote'>('local')
   const [isCreating, setIsCreating] = React.useState(false)
   const [error, setError] = React.useState<string | null>(null)
-  const [remoteRunsEnabled, setRemoteRunsEnabled] = React.useState(false)
-
-  React.useEffect(() => {
-    if (typeof window !== 'undefined') {
-      const flagValue = window.localStorage.getItem('remote-runs')
-      setRemoteRunsEnabled(flagValue === 'true')
-    }
-  }, [])
+  
+  // Check if remote runs feature is enabled via environment variable
+  const remoteRunsEnabled = import.meta.env.VITE_REMOTE_RUNS_ENABLED === 'true'
 
   const handleCreate = async () => {
     if (!unitName.trim()) return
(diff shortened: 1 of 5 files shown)


taco/internal/auth/signed_url.go
```diff
@@ -65,4 +66,84 @@ func VerifySignedUrl(signedUrl string) error {
...
+	
+	return isValid
+}
+
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
 }
```
motatoes: any reason we are implementing our own min function ?

ui/src/components/UnitCreateForm.tsx
```diff
@@ -120,10 +123,11 @@ export default function UnitCreateForm({
...
           <label
             htmlFor="unit-type-remote"
-            className={`relative flex cursor-not-allowed items-start gap-4 rounded-lg border p-4 md:p-5 opacity-60 bg-muted/30`}
+            className={`relative flex cursor-pointer items-start gap-4 rounded-lg border p-4 md:p-5 transition-colors hover:bg-muted/50 ${unitType === 'remote' ? 'ring-2 ring-primary border-primary' : 'border-muted'}`}
+            onClick={() => setUnitType('remote')}
           >
-            <RadioGroupItem id="unit-type-remote" value="remote" disabled className="sr-only" />
-            <div className="flex h-10 w-10 items-center justify-center rounded-md bg-muted text-muted-foreground">
+            <RadioGroupItem id="unit-type-remote" value="remote" className="sr-only" />
+            <div className="flex h-10 w-10 items-center justify-center rounded-md bg-primary/10 text-primary">
```
motatoes: we don't want to go live with this yet, should keep button disabled

motatoes: or a quick feature flag

motatoes (approved):

ui/src/components/UnitCreateForm.tsx
breardon2011: used an envar on backend/frontend so we can control tightly until sandboxed

taco/internal/auth/signed_url.go
breardon2011: good catch

PR #2466, opened 2025-11-20
Switch from callbacks to webhooks for repos

ZIJ:

> [!IMPORTANT]  

---

- [ ] I did **not** use AI tools in this contribution.
- [x] I used AI tools and have disclosed details below.

**Details (if applicable):**
> used codex heavily

---

#### 💡 Notes

### DIFF
--- a/backend/controllers/github_installation.go
+++ b/backend/controllers/github_installation.go
@@ -2,12 +2,15 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
+	"time"
 
 	"github.com/diggerhq/digger/backend/models"
 	"github.com/diggerhq/digger/backend/utils"
 	"github.com/google/go-github/v61/github"
+	"github.com/sethvargo/go-retry"
 )
 
 func getAccountDetails(account *github.User) (string, int64) {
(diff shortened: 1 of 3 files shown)


backend/controllers/github_installation.go
```diff
@@ -20,30 +128,140 @@ func handleInstallationDeletedEvent(installation *github.InstallationEvent, appI
...
+func handleInstallationUpsertEvent(ctx context.Context, gh utils.GithubClientProvider, installation *github.InstallationEvent, appId int64) error {
+	installationId := installation.Installation.GetID()
+	appIdFromPayload := appId
+	if installation.Installation.AppID != nil {
+		appIdFromPayload = installation.Installation.GetAppID()
+	}
+
+	accountLogin, accountId := getAccountDetails(installation.Installation.Account)
+
+	link, err := models.DB.GetGithubInstallationLinkForInstallationId(installationId)
```
motatoes: in case its first instllation the link will only be created on the callback and here we depend on it so we need a retry mechanism incase the webhook arrives before the callback over here. We can retry like 3 to 5 times to fetch the link and if not we call it a failure

motatoes (commented):
In the callback function there is this part on top

```	codeParams, codeExists := c.Request.URL.Query()["code"]
	if !codeExists || len(codeParams) == 0 {
		slog.Error("There was no code in the url query parameters")
		c.String(http.StatusBadRequest, "could not find the code query parameter for github app")
		return
	}
```

However since moving updates to webhook a missing code parameter is no longer considered an error (the update will happen in the webhook) so that step should return a success message

motatoes (approved):

backend/controllers/github_installation.go
ZIJ: addressed

PR #2492, opened 2025-11-27
Move from callbacks to webhooks for repos handling - again - on top of v0.6.136 this time

ZIJ:

> CC using the diff from #2466


ZIJ:
Merging as it seems to be stable after 1d in production released from branch

PR #2503, opened 2025-12-04
introduce monitoring

motatoes:
mainly introduction of structured logging in the ui to log

- status code response
- latency
- method
- path

so we can have metrics around that

used cursor's help to get the latencies and status codes


breardon2011 (approved):
LGTM

PR #2506, opened 2025-12-05
use goroutines during checks to fix the latency issue

motatoes:
we have spiky latencies in the "set-status" endpoint. On inspection of logs I found that the biggest factor is the summary generator service. To avoid huge latency in response and http timeout instead


breardon2011 (approved):
Approved

PR #2508, opened 2025-12-09
overide get_env function to avoid generation errors

motatoes:
Since terragrunt generation and parsing happens on the backend, the function get_env('XYZ") in any terragrunt module would throw an exception:

Screenshot 2025-12-09 at 11 32 39

This is solved by overriding that function during generation as well

* No ai used


breardon2011 (approved):
approved

PR #2509, opened 2025-12-09
handle dev mode

motatoes:
After recent changes to log everything in UI it seems that dev mode was broken, page would load but then would crash with this error:

```
Failed to fetch dynamically imported module: https://mo-digger-test.ngrok.app/src/routes/_authenticated/_dashboard.tsx?tsr-split=component
```

The fix is to detect internal and dynamic component requests, then avoid interrupting them during the logging

* Used cursor to direct this change


breardon2011 (approved):
approved

PR #2513, opened 2025-12-12
Add an option to force run workflows from the repo default branch

languitar:
When using OIDC to obtain cloud credentials in workflows, using the digger workflow from the PR branch might be a security issue. Malicious actors could modify the workflow or add any other workflow to a PR to obtain cloud credentials without any way for code owner approvals preventing this. By being able to force the use of the digger workflow from the repository's main branch, the cloud roles can be configured to only trust runs on that branch, which can in turn be secured using typical PR approvals. That way, malicious workflows could only end up there after having passed a review.

It is not an option to implement this option in the digger CLI / digger.yml. This could then be modified in a malicious PR.

Potential discussion points:

* Should this option be processed earlier and not late in the github backend?

Code generated by Copilot; modified, reviewed and verified manually.

Recreation of #2497 with a proper branch

### DIFF
--- a/backend/ci_backends/github_actions.go
+++ b/backend/ci_backends/github_actions.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log/slog"
 
+	"github.com/diggerhq/digger/backend/config"
 	"github.com/diggerhq/digger/backend/utils"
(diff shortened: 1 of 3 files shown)


languitar:
@motatoes any chance to get this into digger? We'd like to get rid of our custom fork of the backend.

PR #2538, opened 2026-01-15
support for apply policies

motatoes:
Introduction of apply polices which are applied just before an apply goes through. It takes the  same input as the plan policy. Apply policy will also pass the plan summary **IF** plan artefacts are configured, otherwise it will not pass it.

Ai declaration: used claude code in a directed way to ship this feature


breardon2011 (approved):
approved

PR #2539, opened 2026-01-16
improve error messages for units when permission denied

motatoes:
Got this error message on unit creation for an old org, logs showed that it was actually a forbidden message so we have the fix for it now to properly show the error message in that case:

Screenshot 2026-01-15 at 17 04 30


breardon2011 (approved):
Approve

Issue #2541, opened 2026-01-19
Digger.yml reference seems to be out of date


nis-thac:
The documentation for [digger.yml](https://docs.opentaco.dev/ce/reference/digger.yml) seems to be out of date. For example, in the [PR noise reduction](https://docs.opentaco.dev/ce/howto/noise-reduction) docs, it references a 'reporting' key somewhere in the digger.yml, which does not exist in the reference.

emersonf-cat:
Same issue re: `disable_digger_apply_comment` and `disable_digger_apply_status_check` referenced in the [Apply on Merge](https://docs.opentaco.dev/ce/howto/apply-on-merge#apply-on-merge) page.

Please create a JSON Schema that captures the valid configuration of `digger.yml`, creating a source of truth. It should contain a description of each field as well as any validation rules. You can generate documentation from it, and IDEs can verify a configuration and flag issues early.

s1ntaxe770r:
Hey there, 


Thanks for opening this up, looking into it

nis-thac:
Thanks for your PR. The new reference is a lot better! Sadly, the current docs still aren't accurate. The one I have found so far is `comment_render_mode`. The [docs](https://docs.opentaco.dev/ce/reference/digger.yml#param-comment-render-mode) say allowed values are basic and detailed. When executing with detailed, I however get the following error:

❌ Error loading digger config: error loading digger.yml: error cloning and loading config: invalid value for comment_render_mode, detailed expecting basic, group_by_module

s1ntaxe770r:
Ah gotcha, let me take another look into it

joshuamkite-nfb:
Created a new PR for this: https://github.com/diggerhq/digger/pull/2566 to:

- auto-generate schema from the code itself
- validate said schema
- test existing schema against code base in CI to combat drift
- add schema produced to docs

PR #2547, opened 2026-01-22
Docs: Schema generation

s1ntaxe770r:
This pr adds a `schema` sub command to enable `dgctl` generate a  jsonschema for `digger.yml` config files , as requested in #2541
 
> _Example: Used ChatGPT to help with doc phrasing._  
> _Example: Code generated by Copilot; reviewed and verified manually._

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@
   <a href="https://docs.digger.dev/">Docs</a>
 </h2>
 
-
 ---
 
 > Heads-up: We’ve rebranded!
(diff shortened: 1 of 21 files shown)


emersonf-cat:
@s1ntaxe770r to clarify, my suggestion was to be "contract first". In other words, creating and maintaining a JSON Schema file **by hand** for the configuration, which is then used to generate code and documentation. Not generating the schema from documentation or code.

JSON Schema will help you be deliberate about configuration files and their changes, and with appropriate tooling, keep your documentation always in sync. It also lets you implement commands like `validate` to check if Digger config files are valid, and failing fast if they're not.

motatoes:
Hello @emersonf-cat thanks for your suggestion! Generating schema from code is our preferred approach because code is the source of truth, it will never drift away from the actual valid schema, it can be versioned and less prone to errors.

motatoes (approved):

emersonf-cat:
@motatoes there are pros and cons both ways, but happy to see the creation of schemas using either approach. 

I took a quick look, and I think you won't currently be able to drive documentation from the schema because the [config structs](https://github.com/diggerhq/digger/blob/develop/libs/digger_config/yaml.go#L10) don't actually contain documentation. If going code-first, you'll need to add the documentation to those structs, [using Go comments](https://github.com/invopop/jsonschema?tab=readme-ov-file#using-go-comments) or using the `jsonschema: ` and `jsonschema_description:` struct tags shown [in this example](https://github.com/invopop/jsonschema?tab=readme-ov-file#example).

Looking forward to
- website documentation driven from schemas
- versioned schemas
- hosted schema files, so IDEs can retrieve them from the web and use them

PR #2549, opened 2026-01-23
Add digger:force label to bypass impacted projects limit

motatoes:
## Summary
- Add support for `digger:force` label to bypass the `ImpactedProjectsPerChange` limit
- Allow users to intentionally proceed with plan/apply when impacting more projects than the configured limit
- Add helpful error messages informing users about the bypass option

## Problem
When a PR impacts more projects than the `DIGGER_MAX_PROJECTS_PER_CHANGE` limit (default: 100), Digger blocks the operation to protect against GitHub API rate limits. However, there are legitimate cases where users need to bypass this limit for intentional large-scale changes.

## Solution
Introduced a `digger:force` label that users can add to their PRs to bypass the limit check. When this label is present:
- The impacted projects limit is not enforced
- A warning log is created for audit purposes
- Operations proceed normally

## Changes
- ✅ Modified `backend/controllers/github_comment.go` to check for `digger:force` label before enforcing limit
- ✅ Modified `backend/controllers/github_pull_request.go` to check for `digger:force` label before enforcing limit
- ✅ Updated error messages to inform users they can add the label to bypass the limit
- ✅ Added warning logs when limit is bypassed for audit trail

## Test plan
- [ ] Test normal flow: PR with > limit projects without label should fail with updated error message
- [ ] Test bypass flow: PR with > limit projects with `digger:force` label should proceed
- [ ] Verify warning logs are created when bypass is used
- [ ] Test both PR event handler and comment event handler

## Example Usage
When encountering:
```
❌ Error the number impacted projects 421 exceeds Max allowed ImpactedProjectsPerChange: 100
```

Simply add the `digger:force` label to the PR, and Digger will bypass the check and proceed with the operation.

------

AI Delcaration: Used claude code for this PR in a directed way

### DIFF
--- a/backend/controllers/github_comment.go
+++ b/backend/controllers/github_comment.go
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"os"
 	"runtime/debug"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


breardon2011 (approved):
approved

PR #2561, opened 2026-02-02
always infer orgID and userID from session

motatoes:
The tanstack ui layer has been in some cases trusting userID and orgID parsed from some frontend components - sent to server functions. To make this more secure we should instead always parse the JWT token on the server functions and retrieve orgID and user ID from there.

AI Disclosure: I have used claude code to plan and ship this feature in a directed way


breardon2011 (approved):
approved - messaged on other opening

PR #2562, opened 2026-02-03
additional verification of org_id and user_id in token service

motatoes:
The token service is meant to be internal and only callabale from the gateway layer (ui proxy). The calls are authenticated and it learns about who the org_id and user_id making the call based on that. In this we add additional definsible checks on org_id and user_id within the service specifically for create, delete and list tokens - this is to add an extra layer of security in case there are any faults in the ui layer checks.

Ai assistance: used. claude code in a directed manner


breardon2011 (approved):
LGTM

PR #2563, opened 2026-02-03
additional validation in the statesman service

motatoes:
Adding additional checks around unit ID's to ensure that they belong to the current orgID/userID especially during the unit delete operation. Protects against IDOR attacks

AI disclosure: used claude code in a directed way


breardon2011 (approved):
LGTM

PR #2564, opened 2026-02-03
fix: retry session refresh when organizationId is missing

motatoes:
## Summary
- Fixes race condition where new users get `organization_id must be a string` error from WorkOS
- When a user signs up, the webhook that creates their organization may not have processed yet
- The session JWT lacks an `org_id` claim, causing WorkOS API calls to fail

This happens when the webhook arrives after the session is created, which means that the JWT token does not have any orgID claim attached to it. In order to fix this we add a retry mechanism dcuring the withAuth call which attempts to refresh the token and hopefully get the right orgID just in time

Ai disclosure: used claude code in a directed way to implement this change


breardon2011 (approved):
approved

Issue #2542, opened 2026-01-19
Project-Level Check never completes


nis-thac:
When running digger with a single project, the project-level check never completes and has no output. This means that it is impossible to see at a glance in the PR overview whether a plan has completed or not.

<img width="1429" height="735" alt="Image" src="https://github.com/user-attachments/assets/9aa16cba-8136-4118-8aeb-b0120f8ba391" />

//Edit: Together with

```yaml
reporting:
  comments_enabled: false
```

it is impossible to see the plans.

nis-thac:
The error seems to be that the checks have a character limit of 65535, but my plans keep exceeding it.

```json
{
    "time": "2026-01-21T13:00:00",
    "severity": "ERROR",
    "msg": "Failed to update check run",
    "app": "digger-backend",
    "inputCheckRunId": "123456789",
    "error": "PATCH https://api.github.com/repos/ORG/REPO/check-runs/123456789: 422 Invalid request.\n\nOnly 65535 characters are allowed; 85471 were supplied. []"
}
```

s1ntaxe770r:
Hey @nis-thac, 

Thanks for sending this in, adding in the JSON also helped a lot.

As a follow up i was curious how large your terraform project is and how i might reproduce this ?

nis-thac:
Hey @s1ntaxe770r,
Our project has approximately 800 resources. This isn't as important tho, since the character limit can easily be hit with fewer if the output of the plan is too large. I see you have already created a PR, thanks a lot!

Now there is also a [limit for the length of a GitHub comment](https://github.com/dead-claudia/github-limits?tab=readme-ov-file#issue-comments). Since we don't want to be in a situation where we cannot see a plan, maybe you could consider a second pr to attach the plan as a file to the run logs if they don't fit.

sav-valerio:
Hi @s1ntaxe770r,

We've noticed a side case with your changes on https://github.com/diggerhq/digger/pull/2550 that also seem to block auto-merge for us.

Please have a look at https://github.com/diggerhq/digger/pull/2615 and let me know if you think it makes sense.

Thanks!

s1ntaxe770r:
Taking a look now! And thanks for bringing this up

s1ntaxe770r:
reopening, as GitHub has updated their API

PR #2565, opened 2026-02-05
Feat/singed url

s1ntaxe770r:

Since GitHub enforces a character limit, rather than hacking the api to upload plans as artifacts, this pr adds an endpoint to serve plans via plain text which is in turn returned when a check exceeds the character limit 

Related to: #2542 

> used claude for docs components & finding the singed url package

### DIFF
--- a/backend/bootstrap/main.go
+++ b/backend/bootstrap/main.go
@@ -211,6 +211,9 @@ func Bootstrap(templates embed.FS, diggerController controllers.DiggerController
 
 	admin.POST("/tokens/issue-access-token", controllers.IssueAccessTokenForOrg)
 
+	// Public endpoint for job output - auth via signed URL (no middleware needed)
+	r.GET("/api/jobs/:jobId/output", diggerController.GetJobOutput)
+
 	r.Use(middleware.CORSMiddleware())
 
 	// internal endpoints not meant to be exposed to public and protected behind webhook secret
--- a/backend/bootstrap/main.go
+++ b/backend/bootstrap/main.go
@@ -212,7 +212,7 @@ func Bootstrap(templates embed.FS, diggerController controllers.DiggerController
 	admin.POST("/tokens/issue-access-token", controllers.IssueAccessTokenForOrg)
 
 	// Public endpoint for job output - auth via signed URL (no middleware needed)
-	r.GET("/api/jobs/:jobId/output", diggerController.GetJobOutput)
+	r.GET("/orchestrator/api/jobs/:jobId/output", diggerController.GetJobOutput)
 
 	r.Use(middleware.CORSMiddleware())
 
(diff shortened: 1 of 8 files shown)


backend/utils/signed_url.go
```diff
@@ -0,0 +1,86 @@
...
+// SignURL creates a signed URL valid until expiry.
+// Uses WEBHOOK_SECRET for signing and DIGGER_HOSTNAME as base URL.
+// This duplicates the pattern from taco/internal/auth/signed_url.go for use in the backend.
+func SignURL(path string, expiry time.Time) (string, error) {
+	secret := os.Getenv("WEBHOOK_SECRET")
+	if secret == "" {
+		return "", fmt.Errorf("WEBHOOK_SECRET not configured")
+	}
+
+	baseURL := os.Getenv("DIGGER_HOSTNAME")
```
motatoes: baseUrl is actually "HOSTNAME", i know this is confusing and not so consistent in other services but for all orchestrator examples we use HOSTNAME as the env var. (only exception is for drift service controllers but those are a different container). Eventually we might migrate to DIGGER_HOSTNAME

s1ntaxe770r: Ah gotcha

backend/bootstrap/main.go
```diff
@@ -211,6 +211,9 @@ func Bootstrap(templates embed.FS, diggerController controllers.DiggerController
 
 	admin.POST("/tokens/issue-access-token", controllers.IssueAccessTokenForOrg)
 
+	// Public endpoint for job output - auth via signed URL (no middleware needed)
+	r.GET("/api/jobs/:jobId/output", diggerController.GetJobOutput)
```
motatoes: in the opentaco setup where the UI sits infront of orchestrator and acts as gateway, we need to also expose this endpoint to the ui layer by:

1. adding a gateway route in UI_HOST/orchestrator/api/jobs/:jobID/output -> orchesatrator/api/jobs/:jobID/output (we add orchestrator so that it is namespaced at the gateway level
2. modifying this to be /orchestrator/api/jobs/:jobID/output to match


```suggestion
	r.GET("/orchestrator/api/jobs/:jobId/output", diggerController.GetJobOutput)
```

backend/controllers/projects_helpers.go
```diff
@@ -460,8 +461,19 @@ func UpdateCheckRunForJob(gh utils.GithubClientProvider, job *models.DiggerJob)
 
 	// Character limit check - GitHub check run text field has a 65535 character limit
 	const maxCheckRunTextLength = 65535
-	cutOffMsg := "\n[Character limit exceeded, output truncated]"
 	if utf8.RuneCountInString(job.TerraformOutput) > maxCheckRunTextLength {
+		// Generate signed URL for full output (30-day expiry)
+		path := fmt.Sprintf("/api/jobs/%s/output", job.DiggerJobID)
```
motatoes: To match the above path:

```suggestion
		path := fmt.Sprintf("/orchestrator/api/jobs/%s/output", job.DiggerJobID)
```

PR #2612, opened 2026-03-15
fix(storage): migrate GitHub artifact upload from deprecated v1-v3 API to v4

### LINKED ISSUE #1702: `upload-plan-destination: github` is using deprecated Artifact API scheduled for EOL on November 30, 2024
I noticed the following deprecation warning when using `upload-plan-destination: github`.

> Deprecation notice: v1, v2, and v3 of the artifact actions
The following artifacts were uploaded using a version of actions/upload-artifact that is scheduled for deprecation:  "environments_dev_dir1", "environments_prod_dir2".
Please update your workflow to use v4 of the artifact actions.
Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

image

The minimum reproduction case is as follows:

```.github/workflows/digger_workflow.yml
name: Digger Workflow

on:
  pull_request:
    branches: [ main ]
    types: [ opened, synchronize ]
  issue_comment:
    types: [created]
  workflow_dispatch:

jobs:
  digger-job:
    runs-on: ubuntu-latest
    permissions:
      contents: write      # required to merge PRs
      actions: write       # required for plan persistence
      id-token: write      # required for workload-identity-federation
      pull-requests: write # required to post PR comments
      statuses: write      # required to validate combined PR status

    steps:
      - uses: diggerhq/digger@vLatest
        with:
          no-backend: true
          disable-locking: true
          upload-plan-destination: github
        env:
          GITHUB_CONTEXT: ${{ toJson(github) }}
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

The current vLatest commit ID is b552ebd7556eed0a297af16f44042177c6b, but looking at `action.yml` it doesn't look like we are depending on `actions/upload-artifact`.
https://github.com/diggerhq/digger/blob/b552ebd7556eed0a297af16f44042177c6b755ba/action.yml

Curious, I read the source and found that the current implementation is calling Artifact's API directly.
https://github.com/diggerhq/digger/blob/b552ebd7556eed0a297af16f44042177c6b755ba/libs/storage/plan_storage.go#L32

`api-version=6.0-preview` looked suspicious, so I went to read the actions/upload-artifact implementation and found that `actions/upload-artifact@v4` uses npm's `@actions/artifact` v2.
https://github.com/actions/upload-artifact/commit/0f573694adb2afc7b30f9435fac65e3ad281da4e

At a quick glance `@actions/artifact` v2 endpoints appear to be completely different from `@actions/artifact` v1.
https://github.com/actions/toolkit/blob/%40actions/artifact%401.1.1/packages/artifact/src/internal/utils.ts#L222
https://github.com/actions/toolkit/blob/%40actions/artifact%402.0.1/packages/artifact/src/internal/shared/artifact-twirp-client.ts#L57

As I understand it, even if we are not using `actions/upload-artifact` directly, this old Artifact API will also be likely to unavailable after November 30, 2024, so it looks like something needs to be done before EOL.
https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

kiwamizamurai:
## Summary

Fixes #1702

The old `StorePlanFile` implementation called `_apis/pipelines/workflows/{runID}/artifacts?api-version=6.0-preview` (via `ACTIONS_RUNTIME_URL`), the deprecated Artifacts v1-v3 internal API. GitHub has removed this endpoint, causing HTTP 400 errors for anyone using `upload-plan-destination: github`.

This PR migrates the upload path to the **Artifacts v4 Twirp/JSON protocol** used by `@actions/artifact` v2+, the same protocol that `actions/upload-artifact@v4` uses internally:

1. `POST {ACTIONS_RESULTS_URL}/twirp/github.actions.results.api.v1.ArtifactService/CreateArtifact` → receive `signed_upload_url`
2. `PUT` file contents to the signed Azure Blob URL (`x-ms-blob-type: BlockBlob`)
3. `POST .../FinalizeArtifact` to commit the upload

The download path (`DownloadLatestPlans`, `PlanExists`, `RetrievePlan`) already uses the stable GitHub REST API (`go-github`) and is unchanged.

## Changes

This PR contains three commits, each addressing a distinct piece needed to make v4 upload work:

### 1. `libs/storage/plan_storage.go` — replace v1-v3 calls with v4 Twirp/JSON

The download path is unchanged. Only `StorePlanFile` is rewritten.

### 2. `action.yml` — expose `ACTIONS_RESULTS_URL` to the digger CLI

GitHub Actions does **not** inject the runtime service env vars (`ACTIONS_RUNTIME_TOKEN`, `ACTIONS_RESULTS_URL`, etc.) into shell `run:` steps — they are only injected into Node.js JavaScript actions. The digger CLI is launched from a `run:` step, so it sees these variables as empty.

action.yml already had a Node.js step that re-exports the v1-v3 vars (`ACTIONS_RUNTIME_URL`, `ACTIONS_RUNTIME_TOKEN`, `ACTIONS_CACHE_URL`) via `core.exportVariable()` for exactly this reason. This PR adds `ACTIONS_RESULTS_URL` to that list (one-line change). Without this, the Go code aborts immediately with `ACTIONS_RESULTS_URL is not set`.

### 3. `libs/storage/plan_storage.go` — derive backend IDs from the JWT, not from env vars

The Twirp service rejects requests with `HTTP 401 "invalid auth token"` when `workflow_run_backend_id` is set to `GITHUB_RUN_ID`. The official `@actions/artifact` v2 package does **not** use `GITHUB_RUN_ID` / `GITHUB_RUN_ATTEMPT` — it parses the JWT in `ACTIONS_RUNTIME_TOKEN` and reads the `scp` claim, which has the form:

```
Actions.Results:<workflow_run_backend_id>:<workflow_job_run_backend_id>
```

The IDs the backend ties to the bearer token are baked into the token itself, so they must be extracted from the JWT — not guessed from env vars. This PR adds a small base64-decode-only JWT payload parser (no signature verification — the token is the runner's own) and uses the extracted IDs in `CreateArtifact` and `FinalizeArtifact`.

## Test plan

End-to-end verified with a real PR on a GitHub-hosted runner: https://github.com/kiwamizamurai/digger_tutorial/pull/4

Successful run output:

```
level=INFO msg="Successfully stored plan file in GitHub artifacts"
  owner=kiwamizamurai repo=digger_tutorial artifactName=sandbox size=2607
```

All three Twirp steps succeeded (`CreateArtifact` → signed-URL `PUT` → `FinalizeArtifact`).

- [x] `go build ./libs/storage/...` passes
- [x] `CreateArtifact` returns `signed_upload_url`
- [x] Azure Blob `PUT` succeeds
- [x] `FinalizeArtifact` commits the artifact
- [x] Plan can be downloaded by a subsequent step (existing REST path is untouched)

> Note: Twirp supports both binary protobuf and JSON encoding. This implementation uses JSON to avoid adding a protobuf dependency.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -489,6 +489,7 @@ runs:
           core.exportVariable('ACTIONS_CACHE_URL', process.env['ACTIONS_CACHE_URL'])
           core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env['ACTIONS_RUNTIME_TOKEN'])
           core.exportVariable('ACTIONS_RUNTIME_URL', process.env['ACTIONS_RUNTIME_URL'])
+          core.exportVariable('ACTIONS_RESULTS_URL', process.env['ACTIONS_RESULTS_URL'])
 
     - name: create cache dir
       run: |
--- a/libs/storage/plan_storage.go
+++ b/libs/storage/plan_storage.go
@@ -36,75 +36,78 @@ func (gps *GithubPlanStorage) StorePlanFile(fileContents []byte, artifactName st
 		"size", len(fileContents))
 
 	actionsRuntimeToken := os.Getenv("ACTIONS_RUNTIME_TOKEN")
-	actionsRuntimeURL := os.Getenv("ACTIONS_RUNTIME_URL")
+	actionsResultsURL := os.Getenv("ACTIONS_RESULTS_URL")
 	githubRunID := os.Getenv("GITHUB_RUN_ID")
-	artifactBase := fmt.Sprintf("%s_apis/pipelines/workflows/%s/artifacts?api-version=6.0-preview", actionsRuntimeURL, githubRunID)
+	githubRunAttempt := os.Getenv("GITHUB_RUN_ATTEMPT")
 
-	headers := map[string]string{
-		"Accept":        "application/json;api-version=6.0-preview",
+	if actionsResultsURL == "" {
+		return fmt.Errorf("ACTIONS_RESULTS_URL is not set; GitHub Actions Artifacts v4 requires this environment variable")
+	}
+
+	twirpBase := strings.TrimRight(actionsResultsURL, "/") + "/twirp/github.actions.results.api.v1.ArtifactService"
+
+	jsonHeaders := map[string]string{
 		"Authorization": "Bearer " + actionsRuntimeToken,
 		"Content-Type":  "application/json",
 	}
 
-	// Create Artifact
-	createArtifactURL := artifactBase
-	createArtifactData := map[string]string{"type": "actions_storage", "name": artifactName}
-	createArtifactBody, _ := json.Marshal(createArtifactData)
+	// Step 1: CreateArtifact
+	createReqBody, _ := json.Marshal(map[string]interface{}{
+		"workflow_run_backend_id":     githubRunID,
+		"workflow_job_run_backend_id": githubRunAttempt,
+		"name":                        artifactName,
+		"version":                     4,
+	})
 
-	slog.Debug("Creating GitHub artifact", "url", createArtifactURL, "name", artifactName)
-	createArtifactResponse, err := doRequest("POST", createArtifactURL, headers, createArtifactBody)
-	if createArtifactResponse == nil || err != nil {
-		slog.Error("Failed to create GitHub artifact",
-			"error", err,
-			"artifactName", artifactName)
-		return fmt.Errorf("could not create artifact with github %v", err)
-	}
-	defer createArtifactResponse.Body.Close()
-
-	// Extract Resource URL
-	createArtifactResponseBody, _ := io.ReadAll(createArtifactResponse.Body)
-	var createArtifactResponseMap map[string]interface{}
-	json.Unmarshal(createArtifactResponseBody, &createArtifactResponseMap)
-	resourceURL := createArtifactResponseMap["fileContainerResourceUrl"].(string)
-
-	// Upload Data
-	uploadURL := fmt.Sprintf("%s?itemPath=%s/%s", resourceURL, artifactName, storedPlanFilePath)
-	uploadData := fileContents
-	dataLen := len(uploadData)
-	headers["Content-Type"] = "application/octet-stream"
-	headers["Content-Range"] = fmt.Sprintf("bytes 0-%v/%v", dataLen-1, dataLen)
-
-	slog.Debug("Uploading file to GitHub artifact",
-		"url", uploadURL,
-		"size", dataLen)
-	_, err = doRequest("PUT", uploadURL, headers, uploadData)
+	createURL := twirpBase + "/CreateArtifact"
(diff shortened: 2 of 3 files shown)


s1ntaxe770r:
hey @kiwamizamurai ,

any chance you could link me to a repo where you tested this? or some screenshots

kiwamizamurai:
@s1ntaxe770r sorry I don't have it because I was working on it for long time ago....

s1ntaxe770r:
ahh, gotcha. Mind testing this out? This looks good but i would need to be certain it woks as intended

kiwamizamurai:
@s1ntaxe770r Got it tested end-to-end. Spinning up a real workflow surfaced two more bugs that the original commit didn't account for, so I've pushed two follow-up commits on top of the v4 migration:

**1. `action.yml` — `ACTIONS_RESULTS_URL` wasn't reaching the CLI**

GitHub Actions only injects the runtime service env vars (`ACTIONS_RUNTIME_TOKEN`, `ACTIONS_RESULTS_URL`, …) into Node.js actions, not into shell `run:` steps. The digger CLI runs from a `run:` step and saw the variable as empty — the Go code aborted immediately with `ACTIONS_RESULTS_URL is not set`.

action.yml already had a `core.exportVariable()` step that re-exports the v1-v3 vars via `\$GITHUB_ENV` for this exact reason. I added `ACTIONS_RESULTS_URL` to the same list. One-line fix.

**2. `plan_storage.go` — `workflow_run_backend_id` must come from the JWT, not from `GITHUB_RUN_ID`**

After fixing #1, `CreateArtifact` started returning `HTTP 401 \"invalid auth token\"`. The official `@actions/artifact` v2 doesn't use `GITHUB_RUN_ID` / `GITHUB_RUN_ATTEMPT` for these IDs — it parses the JWT in `ACTIONS_RUNTIME_TOKEN` and reads the \`scp\` claim, which encodes them as:

\`\`\`
Actions.Results:<workflow_run_backend_id>:<workflow_job_run_backend_id>
\`\`\`

The IDs the backend ties to the bearer token are baked into the token itself, so they have to be extracted from there. Added a small base64-decode-only JWT payload parser (no signature check — it's the runner's own token).

**Verified end-to-end on a hosted runner:**

\`\`\`
level=INFO msg=\"Successfully stored plan file in GitHub artifacts\"
  owner=kiwamizamurai repo=digger_tutorial artifactName=sandbox size=2607
\`\`\`

All three Twirp calls succeed (CreateArtifact → signed-URL PUT → FinalizeArtifact). Test repo PR for evidence: https://github.com/kiwamizamurai/digger_tutorial/pull/4

PR description updated to cover all three pieces. Ready for another look 🙏

s1ntaxe770r:
Awesome i would try this out myself , do you have any steps to reproduce myself?

kiwamizamurai:
No? Maybe you can just refer my above pr?

kiwamizamurai:
What blocks this pr?

PR #2619, opened 2026-03-21
conditionally enable summary

### LINKED ISSUE #2613: summaries should be turned off by default
Digger allows for summaries of changes to be made via an LLM, however this an experimental feature which should be turned off by default. This issue seeks to investigate why this isn't the case

s1ntaxe770r:

Used claude to grep for relevant files. The fix was manual 

This pr resolves #2613


s1ntaxe770r:
this might require a closer review cuz i am introducing a check for if `aiSummaryEnabled`

PR #2630, opened 2026-04-02
fix: remove web middleware from github setup and exchange-code routes

kaylenmistry:
These endpoints are proxied through the UI, which handles authentication before forwarding the request. The backend web middleware expects a session cookie that is not present in proxied requests, causing a 403 on `/orchestrator/github/setup`.

This aligns `/github/setup` and `/github/exchange-code` with `/github/callback`, which was already registered outside the auth group for the same reason.


kaylenmistry:
Closing - PR was raised against wrong base branch, will re-raise against develop.

PR #2634, opened 2026-04-05
fix(drift): fix SplitCodeBlocks line duplication and handle plain text messages

### LINKED ISSUE #745: Drift alerts in Slack > 4000 characters break formatting
Option 1 (suggested by the user): upload a snippet instead of message text

Option 2: send summary instead of full plan output. Long text as message is useless anyways.

https://github.com/containerscrew/tftools#tf-summarize

raman1236:
## Summary

Fixes #745 - Drift alerts in Slack > 4000 characters break formatting

### Bugs Fixed

**1. Line duplication in `SplitCodeBlocks`**

When a line caused the accumulated message to exceed Slack's 4000-character limit, the splitting logic appended the triggering line to **both** the current chunk and the start of the next chunk:

```go
// Before (buggy):
res = append(res, part+"\n"+line+"\n```")  // line added to OLD part
part = "```\n" + line                       // line ALSO added to NEW part
```

This caused every line at a split boundary to appear twice in the Slack notification.

**Fix:** Close the current chunk cleanly without the triggering line, then start the new chunk with it:
```go
// After:
res = append(res, part+"\n```")
part = "```\n" + line
```

**2. Messages without code blocks skipped splitting**

The function returned messages with fewer than 2 `` ``` `` markers unsplit, even when they exceeded 4000 characters. This caused the Slack API to reject them.

**Fix:** Remove the early return and add a plain-text splitting path.

**3. `SendErrorNotificationForProject` didn't split messages**

Unlike `SendNotificationForProject`, the error notification method sent messages directly via `SendSlackMessage` without going through `SplitCodeBlocks`. Long error stack traces would be rejected by Slack.

**Fix:** Use the same split-and-loop pattern.

### Testing

Added 2 new unit tests:
- `TestSplitCodeBlocksNoLineDuplication` — verifies no line appears in multiple parts after splitting
- `TestSplitCodeBlocksPlainTextOver4000` — verifies plain text messages > 4000 chars are split

All existing tests continue to pass:
```
=== RUN   TestSlackSplitLargerMessage      --- PASS
=== RUN   TestSlackSmallerMessageNotSplit   --- PASS
=== RUN   TestSplitCodeBlocksNoLineDuplication --- PASS
=== RUN   TestSplitCodeBlocksPlainTextOver4000 --- PASS
```

### DIFF
--- a/cli/pkg/drift/slack.go
+++ b/cli/pkg/drift/slack.go
@@ -14,18 +14,20 @@ type SlackNotification struct {
 func SplitCodeBlocks(message string) []string {
 	var res []string
 
-	if strings.Count(message, "```") < 2 {
-		res = append(res, message)
-		return res
-	}
+	hasCodeBlocks := strings.Count(message, "```") >= 2
 
 	regex := regexp.MustCompile("\n")
 	split := regex.Split(message, -1)
 	part := ""
 	for _, line := range split {
-		if len(part+line) > 4000 {
-			res = append(res, part+"\n"+line+"\n```")
-			part = "```\n" + line
+		if len(part)+len(line)+1 > 4000 {
+			if hasCodeBlocks {
+				res = append(res, part+"\n```")
+				part = "```\n" + line
+			} else {
+				res = append(res, part)
+				part = line
+			}
 		} else {
 			part = part + "\n" + line
 		}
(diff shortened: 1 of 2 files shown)


raman1236:
Friendly ping — this PR fixes the SplitCodeBlocks line duplication bug in drift detection and adds proper handling for plain text messages. Would appreciate a review when you get a chance. Thanks!

Issue #1538, opened 2024-05-31
Digger fails to parse plans provided by OpenTofu


sikha-root:
v0.4.39 resolved executing Digger jobs with OpenTofu, but plans using `tofu` fail with the following error:

> Failed to Run digger plan command. error checking for empty plan: Error while parsing json file: Unable to parse the plan file: invalid character 'c' looking for beginning of value

<details><summary>Further context (log)</summary>
<p>

```
Running command: opentofu [workspace list]
Running command: opentofu [workspace new lab]
/home/runner/work/_temp/x/tofu-bin workspace new lab
Created and switched to workspace "lab"!

You're now on a new, empty workspace. Workspaces isolate their state,
so if you run "tofu plan" OpenTofu will not see any existing state
for this configuration.
Running command: opentofu [plan -out /home/runner/work/tofu-test/tofu-test/example/org-tofu-test-1-example.tfplan -lock-timeout=3m -var-file=lab.tfvars -input=false -no-color -detailed-exitcode]
/home/runner/work/_temp/uuid/tofu-bin plan -out /home/runner/work/tofu-test/tofu-test/example/org-tofu-test-1-example.tfplan -lock-timeout=3m -var-file=lab.tfvars -input=false -no-color -detailed-exitcode

OpenTofu used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create

OpenTofu will perform the following actions:

  # random_password.example will be created
  + resource "random_password" "example" {
      + bcrypt_hash = (sensitive value)
      + id          = (known after apply)
      + length      = 32
      + lower       = true
      + min_lower   = 0
      + min_numeric = 0
      + min_special = 0
      + min_upper   = 0
      + number      = true
      + numeric     = true
      + result      = (sensitive value)
      + special     = false
      + upper       = true
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + default_label_id = "example-dev-example"
  + example_password = (known after apply)
  + test             = "hello"

─────────────────────────────────────────────────────────────────────────────

Saved the plan to:
/home/runner/work/tofu-test/tofu-test/example/org-tofu-test-1-example.tfplan

To perform exactly these actions, run the following command to apply:
    tofu apply "/home/runner/work/tofu-test/tofu-test/example/org-tofu-test-1-example.tfplan"
[...]

PR #2517, opened 2025-12-16
Fix/handle missing digger config gracefully

joshuamkite-nfb:
### Problem
Digger was _sporadically_ showing git cloning errors in repositories that don't have Digger configured:
- "Error loading digger config: error loading digger.yml: error cloning and loading config: git command failed: exit status 128"
- "gnutls_handshake() failed: The TLS connection was non-properly terminated"

### Root Cause
These issue appear to occur when the background projects refresh service and GitHub Actions process repositories without digger.yml files:

1. **Background projects refresh service** clones repositories from the organization
2. **Repositories without digger.yml files** get processed with `generateProjects=true` 
3. **Empty configuration triggers terragrunt parsing** if terragrunt files exist in those repositories
4. **Terragrunt parsing attempts to clone remote repositories** referenced in terragrunt configurations
5. **Network/TLS issues cause git cloning failures**

The `generateProjects=true` parameter is hardcoded in multiple places, e.g.:
- **GitHub Actions**: [`cli/pkg/github/github.go:111`](cli/pkg/github/github.go#L111)
- **Background Projects Refresh Service**: [`background/projects-refresh-service/projects_refesh_main.go:49`](background/projects-refresh-service/projects_refesh_main.go#L49)
- **Backend GitHub Helpers**: [`backend/controllers/github_helpers.go:865-866`](backend/controllers/github_helpers.go#L865-L866)

### Solution
Modified `LoadDiggerConfigYaml` function to handle missing configuration files gracefully by returning an empty configuration instead of an error when no config file is found.

### Changes
- **libs/digger_config/digger_config.go**: Changed error handling to return empty configuration when no config file exists
- **libs/digger_config/converters.go**: Added nil check in `copyWorkflowConfiguration` to prevent panic when workflow config is nil
- **libs/ci/github/github.go**: Added nil pointer safety checks for GitHub payload processing with helper functions `getStringValue` and `getWorkflowCommands`
- **libs/digger_config/digger_config_test.go**: Updated existing test and added new test to verify graceful handling
- **cli/pkg/integration/integration_test.go**: Fixed test JSON to include required sender and repository fields

### Testing
- All existing tests continue to pass (verified with full test suite: `go test -v`)
- New tests verify that missing config files are handled gracefully both with and without project generation
- Existing test coverage ensures repositories with Digger configuration continue to work as expected

### Impact
- Repositories without Digger configuration no longer show git cloning errors
- Maintains full backward compatibility for configured repositories
[...]

PR #2644, opened 2026-04-16
fix(cli): propagate per-project errors from RunJobs to exit message

mwielgosz-esky:
## Problem

We hit exit code 8 in our CI with the message:

```
Failed to run commands. %!s(<nil>)
```

In our case, the underlying failure was `terragrunt show -json` failing during `postProcessPlan()` in a Terragrunt monorepo (~100 projects). The only way to find the actual error was scrolling through thousands of structured log lines looking for the `slog.Error` entry -- the exit message itself gave zero information.

The `%!s(<nil>)` symptom has been seen before (#1538), though that particular instance was caused by OpenTofu JSON parsing (#1545). The error swallowing pattern in `RunJobs()` described below was not addressed.

## Root cause

`RunJobs()` in `cli/pkg/digger/digger.go` logs per-project errors via `slog.Error` but always returns `nil` as its error value:

```go
executorResult, _, err := run(...)
if err != nil {
    slog.Error("error while running command for project", ...)
    appliesPerProject[job.ProjectName] = false
    // error is never collected
}
// ...
return allAppliesSuccess, atLeastOneApply, nil  // always nil
```

The caller in `github.go` formats the nil error with `%s`, producing `%!s(<nil>)`:

```go
usage.ReportErrorAndExit(githubActor, fmt.Sprintf("Failed to run commands. %s", err), 8)
```

## Fix

Track the first error encountered across project runs and return it instead of `nil`:

- Add `var firstErr error` before the job loop
- Capture on first failure: `firstErr = fmt.Errorf("project %v command %v: %w", ...)`
- Return `firstErr` instead of `nil`

This preserves the existing `allAppliesSuccess` flag behavior while giving callers an actionable error message.

## Before / After

- **Before:** `Failed to run commands. %!s(<nil>)` -- no diagnostic value
- **After:** `Failed to run commands. project <name> command <cmd>: <underlying error>` -- identifies which project failed and why

## Regression risk

No behavioral change. All 3 callers (`github.go`, `spec.go`, `manual.go`) already check `!allAppliesSuccess || err != nil`. When `run()` fails, `appliesPerProject` is set to `false`, so `allAppliesSuccess` is already `false` and the failure branch is taken regardless of `err`. The only difference is that `err` now carries the actual message instead of `nil`.

Existing unit tests pass (`go test ./cli/pkg/digger/... -short`).

### :brain: AI disclosure

This PR was authored by an AI agent (Claude/Roo) with human review. The bug was identified by tracing the code path from CI logs showing the `%!s(<nil>)` pattern through `github.go` -> `digger.go` -> `execution.go` -> `terragrunt.go`.

### DIFF
--- a/cli/pkg/digger/digger.go
+++ b/cli/pkg/digger/digger.go
@@ -74,6 +74,7 @@ func RunJobs(jobs []orchestrator.Job, prService ci.PullRequestService, orgServic
 
 	exectorResults := make([]execution.DiggerExecutorResult, len(jobs))
 	appliesPerProject := make(map[string]bool)
+	var firstErr error
 
 	for i, job := range jobs {
 		splits := strings.Split(job.Namespace, "/")
@@ -115,6 +116,9 @@ func RunJobs(jobs []orchestrator.Job, prService ci.PullRequestService, orgServic
 			if err != nil {
 				slog.Error("error while running command for project", "command", command, "projectname", job.ProjectName, "error", err)
 				appliesPerProject[job.ProjectName] = false
+				if firstErr == nil {
+					firstErr = fmt.Errorf("project %v command %v: %w", job.ProjectName, command, err)
+				}
 				if executorResult != nil {
 					exectorResults[i] = *executorResult
 				}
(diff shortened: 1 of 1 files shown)


AndrzejOlender (approved):

s1ntaxe770r:
should be good to go once ci passes

mwielgosz-esky:
Hey @s1ntaxe770r, thanks for the review!

Just a heads up -- the CLI Build check passed on a re-run (completed Apr 24, 18:02 UTC). The failing checks are all `build-and-push` jobs (backend-ee, taco-statesman, drift, ui, token-service, projects-refresh) which need Docker registry secrets unavailable on fork PRs.

This is the same pattern as other recently merged fork PRs with Go/action.yml changes: #2584, #2554, #2517.

Let me know if there's anything else needed on my side.

PR #2654, opened 2026-05-25
fix: deduplicate DependencyProjects before graph insertion to prevent "edge already exists" error

### LINKED ISSUE #2628: Digger fails to parse Terragrunt configurations with cascaded dependencies causing "edge already exists" error
* Terragrunt parsing with `cascadeDependencies: true` and `dependsOnOrdering: true`
* Projects that have both direct and transitive dependencies on the same resource

```
{"severity":"ERROR","msg":"failed to convert YAML to config","error":"failed to create project dependency graph: edge already exists"}
{"severity":"ERROR","msg":"Error loading and parsing Digger config","error":"failed to create project dependency graph: edge already exists"}
```

## Root Cause

When `cascadeDependencies: true` is enabled in `digger.yml`, the Terragrunt parser recursively collects all transitive dependencies. This creates **duplicate entries** in a project's `DependencyProjects` array when:

1. Module A depends directly on Resource X
2. Module A also depends on Module B
3. Module B also depends on Resource X

### Example Scenario

**Terragrunt Structure:**
```
infra/gcp/
├── _global/groups/admin-group/terragrunt.hcl       # Group resource
├── dev/my-project/terragrunt.hcl                   # Project (depends on admin-group)
└── dev/my-project/iam/permissions/secret-access/   # IAM binding (depends on both)
    └── terragrunt.hcl
```

**Dependency Graph from Terragrunt:**
```
digraph {
    "_global/groups/admin-group" ;
    "dev/my-project" ;
    "dev/my-project" -> "_global/groups/admin-group";
    "dev/my-project/iam/permissions/secret-access" ;
    "dev/my-project/iam/permissions/secret-access" -> "_global/groups/admin-group";
    "dev/my-project/iam/permissions/secret-access" -> "dev/my-project";
}
```

### Code Location

The error occurs in `libs/digger_config/converters.go` at line 434:
```go
err = g.AddEdge(dependency, project.Name)
if err != nil {
    return nil, err  // Fails here when duplicate edge is added
}
```

## Reproduction Steps

See `/docs/issue-example/` directory for complete mock Terragrunt files.

1. Create Terragrunt structure with:
   - A shared resource (e.g., IAM group)
   - A parent module that depends on the shared resource
   - A child module that depends on BOTH the parent AND the shared resource

2. Configure `digger.yml` with:
   ```yaml
   generate_projects:
     blocks:
       - block_name: example
         terragrunt: true
         root_dir: "infra/"
         terragrunt_parsing:
           cascadeDependencies: true    # Required to trigger
           dependsOnOrdering: true      # Required to trigger
   ```

3. Open a PR that modifies the child module
4. Observe "edge already exists" error

### Approach: Deduplicate Dependencies Before Graph Creation

**Location:** `libs/digger_config/converters.go`

**Changes:**

1. Add helper function to deduplicate string slices:
   ```go
   func uniqueStrings(str []string) []string {
       keys := make(map[string]bool)
       list := []string{}
       for _, entry := range str {
           if _, value := keys[entry]; !value {
               keys[entry] = true
               list = append(list, entry)
           }
       }
       return list
   }
   ```

2. Deduplicate `DependencyProjects` in `CreateProjectDependencyGraph()`:
   ```go
   projectsMap := make(map[string]Project)
   for _, project := range projects {
       // Deduplicate to prevent "edge already exists" errors
       project.DependencyProjects = uniqueStrings(project.DependencyProjects)
       projectsMap[project.Name] = project
   }
   ```

ramkalapala:
Fixes #2628

## Problem
When `cascadeDependencies: true` and `dependsOnOrdering: true` are both
enabled, recursive dependency expansion adds the same project into
`DependencyProjects` multiple times — once directly and once transitively.

For example with A → X, A → B, B → X:
- X appears in A's `DependencyProjects` twice
- `g.AddEdge(X, A)` is called twice
- The graph library returns `"edge already exists"` and the scan fails

## Root Cause
`CreateProjectDependencyGraph` in `libs/digger_config/converters.go`
iterates `DependencyProjects` and calls `g.AddEdge()` for each entry
with no duplicate check.

## Fix
Add a `seen` map inside the dependency loop to skip duplicate entries
before they reach `g.AddEdge()`. This is a minimal, safe change — no
new imports required, no behaviour change for non-duplicate cases.

## Testing
Existing test suite unchanged. The fix only affects cases where
`DependencyProjects` contains duplicates.

### DIFF
--- a/libs/digger_config/converters.go
+++ b/libs/digger_config/converters.go
@@ -404,7 +404,12 @@ func CreateProjectDependencyGraph(projects []Project) (graph.Graph[string, Proje
 				return nil, err
 			}
 		}
+		seen := make(map[string]bool)
 		for _, dependency := range project.DependencyProjects {
+			if seen[dependency] {
+				continue
+			}
+			seen[dependency] = true
 			_, err := g.Vertex(dependency)
 
 			if errors.Is(err, graph.ErrVertexNotFound) {


ramkalapala:
Hi team, raised this PR to address #2628. @s1ntaxe770r, would really appreciate a review and approval when you get a chance.

PR #2661, opened 2026-06-11
fix: allow digger apply when digger/apply is the sole required blocking check

joshuamkite-nfb:
## Current issue

When `digger/apply` is configured as a required GitHub branch protection check, `IsMergeable()` returns `false` before the first apply runs — the PR is in state `blocked` because the required check hasn't passed yet. This prevents apply from ever executing, making it impossible to satisfy the check. See #1180.

## Fix

When the PR state is `blocked`, inspect the check runs on the head commit. If `digger/apply` checks are the only non-passing checks, allow the apply to proceed. All other blocking conditions (failing CI, missing approvals, diverged branch) continue to prevent apply as before.

**Changes:** `libs/ci/github/github.go` only — `IsMergeable()` delegates to a new `isBlockedOnlyByDiggerApply()` helper when state is `blocked`. No config changes, no other VCS providers affected.

## Enabled by this fix

Once this patch is deployed, `digger/apply` can be added as a required branch protection check on IaC repos. This enforces that apply has successfully completed before any PR can be merged — PRs where apply has never been run will be blocked at the merge button.

---

### :brain: **Ai UsageDetails (if applicable):**

This PR was written primarily by Claude Code (claude-sonnet-4-6) with the fix direction derived from the issue description in #1180. The implementation was reviewed and verified by the author.

Issue #1352, opened 2024-04-19
top issues


motatoes:

MrKeiKun:
is hacktoberfest up for those issue? :) i still have time today till this weekend

EDIT:

11 and 33 is annoying btw.

PR #2672, opened 2026-07-16
fix: isBlockedOnlyByDiggerApply should honor required-check status, not just success conclusions

### LINKED ISSUE #1180: digger apply mergability check should not fail if the "digger/apply" status check is the only thing blocking
In #1150  we have implemented an aggregate status check for digger `digger/plan` and `digger/apply`. Although those are set correctly there is an additional mergability check which causes digger apply step to fail IF the PR is not mergable.

This has introduced chicken-egg situation where you want digger/apply check to succeed for PR to be mergable but digger/apply needs to complete

We should do an additional check on mergability where if PR state is "blocked" and the only check blocking it is "digger/apply" then to still proceed with the apply

Screenshot 2024-02-19 at 12 10 54

joshuamkite-nfb:
**Fixes #1180** - # 7 on the community-ranked [Top Issues list](https://github.com/diggerhq/digger/issues/1352).

## Problem

`isBlockedOnlyByDiggerApply` (added in #2661) is meant to let `digger apply` proceed when a
PR's `mergeable_state` is `"blocked"` only because `digger/apply` itself hasn't passed yet.
Trying to productionise this against a repo with several other required checks turned up two
remaining bugs:

1. It only accepted a check-run conclusion of exactly `"success"`. GitHub's docs state
   required checks accept `"success"`, `"skipped"`, or `"neutral"` - a required check
   concluding `neutral` (a valid pass) caused Digger to treat the PR as blocked by more than
   `digger/apply`, and refuse to proceed.
2. It fetched every check-run ever posted to the commit via REST
   (`GET /commits/{sha}/check-runs`), with no way to tell which ones are actually required. A
   calling workflow's own native job-status check, or Digger's own project-scoped check (e.g.
   `myproject/apply`, distinct from the generic `digger/apply`), could each incorrectly block
   apply even though neither is required for merging.

## What this does

- Replaces the REST check-runs list with the PR's `statusCheckRollup` via GraphQL, which
  carries a per-context `isRequired(pullRequestNumber: ...)` flag - the same rolled-up
  merge-box view any collaborator can see, so it doesn't need the branch-protection admin
  scope that `GET /branches/{branch}/protection` or GraphQL's `branchProtectionRules` require
  (confirmed `403`/`FORBIDDEN` for a standard token on both). Only required contexts are now
  considered when deciding if a PR is blocked.
- Accepts `SUCCESS`, `SKIPPED`, and `NEUTRAL` as passing states, matching [GitHub's documented
  required-check spec](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches)
- Extracts the decision logic into a pure, unit-testable function, `blockedOnlyByDiggerApply`.
- `GetCheckRunsForCommit` (REST) is untouched - it's still used elsewhere for check-run
  *updates*, a different concern.
- Docs:
  - Clarified `disable_digger_apply_status_check`'s interaction with apply-on-comment vs.
    apply-after-merge - it can leave a required `digger/apply` check stuck `in_progress`
    forever if left enabled outside the apply-after-merge flow. Added example config for both
    flows.
  - Added a walkthrough for actually configuring `digger/plan`/`digger/apply` as required
    status checks in GitHub branch protection.
  - Documented a gotcha where a required check pinned to a specific GitHub App ID silently
    blocks merges after a self-hosted backend re-installs its App (e.g. a blue/green deploy).
  - Added the `checks: read` permission (needed to read `statusCheckRollup`) to every example
    workflow permissions block that was missing it.

## Testing

- **Verified via a real deployment**, not just unit tests: built this fix into a backend image
  and deployed it to an actual self-hosted Digger backend. The CLI's prebuilt-binary download
  is hardcoded to fetch from `diggerhq/digger`'s own releases regardless of which repo the
  action is checked out from, so to test the CLI side without patching `action.yml`, the
  calling workflow built the CLI from source from this fork's branch and ran it via the
  existing `local-dev-mode`/`local-dev-cli-path` inputs.
- Exercised end-to-end against live PRs with several other required checks (security
  scanners, a metadata check) plus `digger/apply` itself, with `digger/apply` configured as a
  required branch-protection check: approving the PR and commenting `digger apply` correctly
  dispatched the job and the apply succeeded, creating a real resource - previously this was
  refused before any workflow even ran.
- 5 new unit tests for `blockedOnlyByDiggerApply`, including a case reproducing a
  non-required check failing (a calling workflow's own job-status check and a project-scoped
  check) to confirm it's correctly ignored.
- Not tested: GitHub Enterprise Server. The GraphQL endpoint URL construction should work the
  same way there but hasn't been verified against a real GHES instance.

This PR - the code fix, tests, and documentation changes - was written with assistance from Claude
Code, based on my own investigation of the bug and test setup. I reviewed and verified all
generated code and documentation manually, and validated the fix via real deployments with a
self-hosted Digger backend (see Testing above).

### DIFF
--- a/docs/ce/cloud-providers/authenticating-with-oidc-on-aws.mdx
+++ b/docs/ce/cloud-providers/authenticating-with-oidc-on-aws.mdx
@@ -26,6 +26,7 @@ jobs:
       pull-requests: write # required to post PR comments
       issues: read         # required to check if PR number is an issue or not
       statuses: write      # required to validate combined PR status
+      checks: read         # required to check mergeability if digger/apply is a required status check
 
     steps:
       - name: digger run
--- a/docs/ce/getting-started/github-actions-+-aws.mdx
+++ b/docs/ce/getting-started/github-actions-+-aws.mdx
@@ -70,6 +70,7 @@ jobs:
       pull-requests: write # required to post PR comments
       issues: read         # required to check if PR number is an issue or not
       statuses: write      # required to validate combined PR status
+      checks: read         # required to check mergeability if digger/apply is a required status check
 
     steps:
       - uses: actions/checkout@v4
--- a/docs/ce/getting-started/github-actions-and-gcp.mdx
+++ b/docs/ce/getting-started/github-actions-and-gcp.mdx
@@ -70,6 +70,7 @@ jobs:
       pull-requests: write # required to post PR comments
       issues: read         # required to check if PR number is an issue or not
       statuses: write      # required to validate combined PR status
+      checks: read         # required to check mergeability if digger/apply is a required status check
     steps:
     - uses: actions/checkout@v4
     - name: ${{ fromJSON(github.event.inputs.spec).job_id }}
--- a/docs/ce/getting-started/with-opentofu.mdx
+++ b/docs/ce/getting-started/with-opentofu.mdx
@@ -90,6 +90,7 @@ Place it at `.github/workflows/digger_workflow.yml` (name is important!)
           pull-requests: write # required to post PR comments
           issues: read         # required to check if PR number is an issue or not
           statuses: write      # required to validate combined PR status
+          checks: read         # required to check mergeability if digger/apply is a required status check
 
         steps:
           - uses: actions/checkout@v4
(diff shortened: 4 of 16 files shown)


nis-thac (approved):
this works for us

PR #2701, opened 2026-08-23
fix(drift): surface failed drift checks instead of masquerading as "no drift"

AR-2910:
## Problem
When a drift-check job fails (plan/init error), the project's drift status is left untouched — typically `no drift`. A broken drift check is indistinguishable from a clean one, so real drift can go unnoticed for as long as the check keeps failing (we hit exactly this in production: a regression left a subset of projects permanently "clean" while their checks were erroring).

## Change
- New `DriftStatus` value **`check failed`**, set when a drift job's status callback reports failure. The error output is stored in `DriftTerraformPlan` so the UI shows *why*, and `LatestDriftCheck` is stamped.
- The stale to-create/update/delete counts are **cleared** on failure so the next successful run re-derives drift from a clean baseline (`wasEmptyPlan == true`) and correctly transitions to either `no drift` or `new drift` — without this a still-drifting project would stay stuck in `check failed`.
- Drift Slack notification renders check-failed projects as a :red_circle: row.
- `execution.Plan()` now returns the captured stdout/stderr on error (was `"", ""`), so the failure reason is actually available to store.

## Testing
- `TestProjectDriftCheckFailedRecordsFailureAndRecovers` (new, uses the existing sqlite test harness): asserts the failure recording and both recovery paths.
- Full `backend/controllers` test suite passes; `backend`, `drift`, and `libs` all build.
- This change has been running in our production fork since 2026-08-06 across a ~140-project nightly drift fleet; the failure state fires on real plan errors and recovers cleanly on the next successful check.

Per the contributing guide: this change was implemented with AI assistance (Claude Code) in our production fork, then human-reviewed, tested, and production-validated before being extracted for upstream.

### DIFF
--- a/backend/controllers/projects.go
+++ b/backend/controllers/projects.go
@@ -990,6 +990,20 @@ func (d DiggerController) SetJobStatusForProject(c *gin.Context) {
 			"batchId", batchId,
 		)
 
+		// Record drift-check failure so a failed check is distinguishable from "no drift"
+		// (otherwise the project silently stays at its default "no drift" / never-checked).
+		if isDriftJob, dErr := IsDriftStatusJob(job); dErr != nil {
+			slog.Warn("Could not determine if failed job is a drift job", "jobId", jobId, "error", dErr)
+		} else if isDriftJob {
+			if project, pErr := models.DB.GetProjectByName(orgId, job.Batch.RepoFullName, job.ProjectName); pErr != nil {
+				slog.Warn("Could not load project for drift check-failed update",
+					"jobId", jobId, "projectName", job.ProjectName, "error", pErr)
+			} else if sErr := ProjectDriftCheckFailed(*project, request.TerraformOutput); sErr != nil {
+				slog.Warn("Could not update project drift check-failed state",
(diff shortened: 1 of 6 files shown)


AR-2910:
Following up on this one — two points that might come up in review, addressed preemptively:

**Why `Save` rather than a scoped `Updates`:** `ProjectDriftCheckFailed` deliberately mirrors the adjacent `ProjectDriftStateMachineApply`, which persists via the same `Save(&project)` on a freshly-loaded row, so the drift-persistence code paths stay consistent. Happy to switch to a column-scoped `Updates` if you'd prefer the narrower write.

**No migration needed:** `DriftStatus` is a plain string column (no DB-level enum), so the new `"check failed"` value requires no schema change — existing rows are unaffected and the value round-trips through the existing column.

Happy to iterate on anything else — and #2702 is the companion PR if you're doing a review pass. Thanks!


cc @ZIJ — saw this PR served as a test case for the OpenComputer reviewer; the two points its dry-run surfaced are exactly the ones addressed above. Would appreciate a review pass on this and #2702 whenever you get a chance 🙏

PR #2710, opened 2026-09-13
fix(github): prevent panic when open issue has no body (#2675)

### LINKED ISSUE #2675: Drift detection panics when an open GitHub issue has no body
## What happened

Digger v0.6.148 can finish a drift plan and then panic before it publishes the GitHub issue. The trigger is an unrelated open issue whose API response contains `"body": null`.

This stops the entire drift run. In the run that exposed the bug, Digger checked two of 15 projects and never reached the remaining thirteen.

## Reproduction

1. Create an open GitHub issue without a body in the repository that runs Digger.
2. Run Digger in `drift-detection` mode with `INPUT_DRIFT_GITHUB_ISSUES=true`.
3. Let any project produce a non-empty plan.
4. Digger tries to create or update the drift issue and panics while listing the repository's existing issues.

The failure occurs with the EE CLI on Linux ARM64 and OpenTofu, but the panic is in the shared GitHub service rather than an architecture- or Terraform-specific path.

## Expected behavior

An issue without a body should be represented with an empty string. Digger should continue looking for the matching drift issue, publish the notification, and scan the remaining projects.

## Actual behavior

```text
Info: Sending drift notification regarding project: artifactory-access
panic: runtime error: invalid memory address or nil pointer dereference

github.com/diggerhq/digger/libs/ci/github.GithubService.ListIssues(...)
    /github/workspace/libs/ci/github/github.go:163

github.com/diggerhq/digger/cli/pkg/drift.(*GithubIssueNotification).SendNotificationForProject(...)
    /github/workspace/cli/pkg/drift/github_issue.go:24
```

The process exits with status 2. No drift issue is created or updated.

## Root cause

[`GithubService.ListIssues`](https://github.com/diggerhq/digger/blob/3053cda26247aa3cb94dcd4b8a253ee3cbc78243/libs/ci/github/github.go#L145-L170) dereferences `issue.Body` for every open issue:

```go
allIssues = append(allIssues, &ci.Issue{
    ID:    int64(*issue.Number),
    Title: *issue.Title,
    Body:  *issue.Body,
})
```

GitHub returns a null body for an issue created without body text, so `*issue.Body` panics. The same dereference is present in v0.6.140, v0.6.146, v0.6.148, and the current `develop` branch.

Using the nil-safe go-github getters would preserve the existing behavior while mapping absent fields to their zero values:

```go
ID:    int64(issue.GetNumber()),
Title: issue.GetTitle(),
Body:  issue.GetBody(),
```

A regression test can mock an issue with a number and title but no body, then assert that `ListIssues` returns it with `Body == ""`.

## AI assistance

Codex helped trace the Actions stack, inspect the affected source, and draft this report. I verified the reproduction and source path.

Tyagiquamar:
Fixes #2675

### Summary of Changes
When GitHub returns an issue with no body (\ody: null\), dereferencing \*issue.Body\ in \GithubService.ListIssues\ causes a runtime panic.

This PR updates \ListIssues\ (and related methods) to use the nil-safe helper getters (\issue.GetNumber()\, \issue.GetTitle()\, \issue.GetBody()\), correctly mapping a \
il\ body to an empty string. A regression unit test \TestListIssuesNilFields\ has also been added to verify this behavior.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

PR #2712, opened 2026-09-14
fix: dispatch cli_release.yml explicitly when recreating vLatest

### LINKED ISSUE #2711: `vLatest` release recreation doesn't trigger CLI binary build — release ships with no assets
## Summary
`vlatest-tag-update.yml` deletes and recreates the `vLatest` release/tag from `develop`, using `actions/github-script` with the default `GITHUB_TOKEN`. GitHub Actions does not let events produced by `GITHUB_TOKEN` trigger other workflow runs, so the `release: released` event this creates never fires `cli_release.yml` (and any other release-triggered build/publish workflows). The recreated release ends up with zero assets.

## Evidence
- Release `vLatest` recreated 2026-09-13T10:17:18Z via `workflow_dispatch` of `vlatest-tag-update.yml`.
- Resulting release: `"assets":[]`.
- No run of `cli_release.yml` (or any other workflow) exists in the time window around that recreation.
- Same-day version-tag releases (e.g. `v0.6.151`, created through the normal release process) triggered `cli_release.yml` and built/uploaded `digger-cli-Linux-X64` successfully.
- Consumers pulling `https://github.com/diggerhq/digger/releases/download/vLatest/digger-cli-Linux-X64` get a 404.

## Impact
Anything depending on `vLatest` for the CLI binary (or other release-triggered artifacts) breaks whenever this workflow is dispatched, since no build/publish step runs against the recreated release.

## Suggested fix
Pass a PAT or GitHub App installation token to the `github-token` input of the `actions/github-script` steps in `vlatest-tag-update.yml` instead of relying on the default `GITHUB_TOKEN`, so the release-publish event it produces cascades into `cli_release.yml` (and equivalent build workflows) as intended.

## Immediate workaround
Manually attach `digger-cli-Linux-X64` (and any other expected assets) to the current `vLatest` release to unblock consumers.

joshuamkite-nfb:
## Summary
`vlatest-tag-update.yml` creates the `vLatest` release via `actions/github-script` using the default `GITHUB_TOKEN`. Releases created that way don't fire the `release` event for other workflows, so `cli_release.yml` never ran and the recreated release shipped with no binary assets attached.

`workflow_dispatch` is exempt from that restriction, so this adds a `workflow_dispatch` trigger to `cli_release.yml` (accepting the tag to build against) and has `vlatest-tag-update.yml` dispatch it explicitly right after creating the release.

Fixes #2711

## Changes
- `cli_release.yml`: add `workflow_dispatch` input `tag_name`; check out and release against that tag (or the release event's tag) instead of only the release event's tag.
- `vlatest-tag-update.yml`: add `actions: write` permission; dispatch `cli_release.yml` for tag `vLatest` after creating the release.

## Test plan
- [ ] Run `vlatest-tag-update.yml` via `workflow_dispatch`
- [ ] Confirm `cli_release.yml` runs and attaches `digger-cli-Linux-X64` to the recreated `vLatest` release


s1ntaxe770r:
Thanks @joshuamkite-nfb for the report (#2711) and the correct diagnosis that releases created with the default GITHUB_TOKEN don't fire the release event for other workflows. We went with the other side of that coin in #2720: create the vLatest release with a fine-grained PAT instead, so the release event fires naturally and all seven asset workflows (cli, cli multiarch, ee cli, dgctl, backend/digger-api, tasks images) run with zero per-workflow changes, exactly like a manually cut v0.6.x release. Dispatching cli_release.yml explicitly would only cover 1 of the 7, and its digger-cli-Linux-X64 asset isn't what action.yml downloads (that's digger-cli-<os>-<arch> from cli_release_multiarch.yml). Closing in favour of #2720.

